PR #40, opened 2014-09-28
Feature: revision excludes

tkruse:
So this is the PR is was preparing. It includes all commits of #38, so #38 should be decided and merged first, and then may this PR be amended.

This PR is mostly for early discussion, though it is fully functional and tested.

The README.md change explain how to use the new feature.

As an example, for a real-world project:

Before
```
Using these dependencies:
project.ext {
    JDK_VERSION = 1.7

    versions = [
            servlet_api        : '3.1.0',
            spring             : '4.0.6.RELEASE',
            spring_boot        : '1.1.5.RELEASE',
            wicket             : '6.16.0',
            jackson            : '2.3.3',
            slf4j              : '1.7.6',
            mockito            : '1.9.5',
            assertj            : '1.6.1'
    ]

}

dependencies {
    compile "org.slf4j:slf4j-api:$versions.slf4j"

    compile("org.apache.commons:commons-lang3:3.3.2") { force = true }

    compile("com.google.guava:guava:18.0") { force = true }

    compile "org.springframework:spring-core:$versions.spring"

    compile("com.fasterxml.jackson.core:jackson-core:$versions.jackson") { force = true }

    compile("org.apache.wicket:wicket-core:$versions.wicket") { force = true }
    compile "org.apache.wicket:wicket-util:$versions.wicket"

    testCompile 'junit:junit:4.11'
    testCompile("org.assertj:assertj-core:$versions.assertj")
    testCompile("org.mockito:mockito-core:$versions.mockito") {
      exclude group: 'org.hamcrest'
    }
}


 - com.google.guava:guava:18.0
 - org.apache.commons:commons-lang3:3.3.2
 - org.assertj:assertj-core:1.6.1

 - com.fasterxml.jackson.core:jackson-core [2.3.3 -> 2.4.2]
 - junit:junit [4.11 -> 4.12-beta-2]
 - org.apache.wicket:wicket-core [6.16.0 -> 7.0.0-M3]
 - org.apache.wicket:wicket-util [6.16.0 -> 7.0.0-M3]
 - org.mockito:mockito-core [1.9.5 -> 1.10.3]
 - org.slf4j:slf4j-api [1.7.6 -> 1.7.7]
 - org.springframework:spring-core [4.0.6.RELEASE -> 4.1.0.RC2]
```

After
```
/ / Using this block:
versionsConfig {
  lookupRevisionMapper = { Dependency dep ->
    if (dep.group == 'org.apache.wicket') {
      return '(,7.0['
    }
    if (dep.group == 'org.springframework') {
      return '(,4.1['
    }
    if (dep.group == 'junit') {
      return '4.12'
    }
  }
}

 - com.google.guava:guava:18.0
 - org.apache.commons:commons-lang3:3.3.2
 - org.assertj:assertj-core:1.6.1

 - com.fasterxml.jackson.core:jackson-core [2.3.3 -> 2.4.2]
 - org.apache.wicket:wicket-core [6.16.0 -> 6.17.0]
 - org.apache.wicket:wicket-util [6.16.0 -> 6.17.0]
 - org.mockito:mockito-core [1.9.5 -> 1.10.3]
 - org.slf4j:slf4j-api [1.7.6 -> 1.7.7]
 - org.springframework:spring-core [4.0.6.RELEASE -> 4.0.7.RELEASE]

 - junit:junit

```
So the manually fixed report shows the update to wicket and spring that matter to me.

Concerns:
- The message "Failed to determine the *latest* version..." does not really fit
- Not sure whether you like the way a the configuration uses a closure to return a String:

An alternative would be to try to mimic the way in default gradle transitive dependencies can be modified:
```
configurations {
    all {
        resolutionStrategy {
            eachDependency(new Action<DependencyResolveDetails>() {
                @Override
                void execute(DependencyResolveDetails details) {
                    def requested = details.requested
                    if (requested.group == 'org.springframework') {
                        details.useVersion "4.0.6"
                    ...
```
See 
http://www.gradle.org/docs/current/userguide/dependency_management.html
section 51.8.2.1
But that would require more code.


ben-manes:
The plugin now uses the configuration's `resolutionStrategy`, so additional exclusions can be specified in the user's build.

tkruse:
Hi Ben,
I fail to see how I can add exclusions to get the result that I want. Can you give an example, maybe for org.springframework:spring-core [4.0.6.RELEASE -> 4.1.0.RC2] instead showing the latest 4.0.x release?

tkruse:
Hm, I got a result by doing this:
```
configurations {
  all {
             componentSelection {
                all { ComponentSelection selection ->
                    if (selection.candidate.group == 'org.mockito' && selection.candidate.version >= '2.0') {
                        selection.reject("ignoring 2.x versions")
                    }
                }
            }
        }
    }
}
```
based on https://docs.gradle.org/current/dsl/org.gradle.api.artifacts.ComponentSelectionRules.html
Is that what you meant?
It is a bit noisy in the logs because all rejected versions yield a log line (twice):
```
Selection of org.mockito:mockito-all:2.0.2-beta rejected by component selection rule: ignoring 2.x versions
Selection of org.mockito:mockito-all:2.0.1-beta rejected by component selection rule: ignoring 2.x versions
Selection of org.mockito:mockito-all:2.0.0-beta rejected by component selection rule: ignoring 2.x versions
Selection of org.mockito:mockito-all:2.0.2-beta rejected by component selection rule: ignoring 2.x versions
Selection of org.mockito:mockito-all:2.0.1-beta rejected by component selection rule: ignoring 2.x versions
Selection of org.mockito:mockito-all:2.0.0-beta rejected by component selection rule: ignoring 2.x versions
```

but I get what I need:
```- org.mockito:mockito-all [1.10.18 -> 1.10.19]```

**Comment by ben-manes** [comment]
Yep, you got it. =)

I think the noise is because (a) multiple repositories being checked and (b) the selection rules being run twice. The logging is in Gradle, so not much we can do about that. I think its only displayed with `--info` though.

The reason the rules are run twice is because they are being run for your current configuration and then again on a copy set to find the latest. For #68 I plan on adding an extension, something similar to yours, so that the selection rule can set only on the dependencyUpdates' copy for evaluation.

**Comment by tkruse** [comment]
It's not pretty in the gradle file. I had thought about writing an extension such that the output would be more verbose, like:
```
- org.mockito:mockito-all [1.10.18 -> 1.10.19 -> 1.11.5 -> 2.1.6]
```
meaning the latest version with same major and minor, the latest with same major, and the global latest. That would serve the same purpose, but go without additional configuration. Not sure if that is easily feasible though, I guess it would involve parsing version numbers, which would result in hacks. 

Just brainstorming in case you see an easy solution one day.

ben-manes:
Oh all suggestions are appreciated! :-)

The rule configuration is very verbose. For the common case of setting a global pattern, like excluding betas, we could have an extension property that internally expands to the rule. For more power users would still have access to the rule to do per component filtering.

The trick for the expanded version list is getting it through the Gradle API. Gradle expects to select an exact version or the latest that meets a criteria. We could have a rule that rejects all versions (except for the current) and captures the version list. This would be a roundabout way to get the information, but would let us show the transition flow. I think this is doable and, like you said, cover most scenarios without any rule configuration required in the user's build file.

Overall I think component selection rules are the right way to go, but it will take a few iterations to use them optimally.

PR #48, opened 2015-01-27
Change file output name

ColtonKaroses:
If I call 'gradle dependencyUpdates -DoutputDir=/any/path/with/permission' and I have a multiple project build, files get overwritten.  This is because the output file is always named 'result.txt', and it is all written to the same folder for each project.  I changed the filenames to be the same as the project name so this doesn't happen.  So now, the output file is located at 'build/dependencyUpdates/exampleProject.txt'.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,11 +58,19 @@ You can also set multiple output formats using comma as the separator:
 gradle dependencyUpdates -Drevision=release -DoutputFormatter=json,xml
 ```
 
-Last, the task property `outputDir` controls the output directory for the report  file(s). The directory will be created if it does not exist.
+The task property `outputDir` controls the output directory for the report  file(s). The directory will be created if it does not exist.
 The default value is set to `build/dependencyUpdates`
 
 ```groovy
 gradle dependencyUpdates -Drevision=release -DoutputFormatter=json -DoutputDir=/any/path/with/permission
+```
+
+Last, the task property 'useProjectAsFilename' controls the filename of the report.  This is a boolean value that defaults to false.
+If the value is false, the output files will all be named 'report' e.g. 'report.xml' or 'report.json'.
+If the value is set to true, the value will be set to the project name. e.g. 'exampleProject.txt'
+
+```groovy
+gradle dependencyUpdates -Drevision=release -DoutputDir=/any/path/with/permission -DuseProjectAsFilename=true
 ```
 
 This displays a report to the console, e.g.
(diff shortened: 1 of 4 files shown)


ben-manes:
This lgtm, but I'd like @jochenberger to weigh in since he's more familiar with this plugin than I am (he did a great job in my absence). I hope to slowly get back involved.

Slightly off topic, I've been wondering if we should have used the Gradle reporting tasks. I haven't looked into them enough to know whether they'd reduce bugs by handling this stuff or have only added complexity.

jochenberger:
I like the idea, but I don't like the fact that the PR changes the current behavior. There's a stable version out whose behavior people might have started to rely on. What about making the file name configurable?

ColtonKaroses:
I have made the file name configurable now.  It should default to the current behavior.

Also, I would be in favor of @ben-manes idea of switching it over to the gradle reporting tasks in the long term.  From my limited experience, switching to reporting tasks handles the output directory issue.

ben-manes:
Should this be a property flag that defines the behavior, or should we make the filename a field on the plugin extension? If we do the latter, then to support this feature would be a configuration DSL that might look like,

```gradle
versionsConfig.filename = project.name
```

The benefit is now the flexibility is in the user's control and they may do strange things we never thought of. The downside is that its another line of code the user might need to add to their build script to customize it.

I think when we hit 1.0 then @jochenberger might agree with us making breaking changes. The current 0.x indicates how initially we don't know how things will evolve, what was needed, or if there would be much adoption. I still wish Gradle added this feature natively...

jochenberger:
Sorry, currently I have very limited time to spend on the plug-in.
I'd prefer that the filename be configurable.
I don't have any experience with the reporting task, but I'll happily review and comment on (and probably also merge) PRs. ;-) And @ben-manes, if you feel like making any breaking changes, just go ahead, it's still your plug-in and *I* am the one who's helping out.

ben-manes:
Thanks @jochenberger, you've been helping out a lot and I know you're busy like all of us. I just don't want you to think I don't appreciate it. =)

The only breaking changes will be in 1.0.

Thunderforge:
@ben-manes Is this pull request to make the file name configurable still a possibility? I would love to be able to do that.

ben-manes:
@Thunderforge I think we'd want to investigate Gradle's reporting features and try to integrate with them if possible. I'm not actively working on this plugin, as its good enough for myself and most users. I can only commit to maintenance work (e.g. Gradle compatibility), though will review contributions if someone is interested in an enhancement.

DDineshkumaragrawal:
Calls going

vanniktech:
@ColtonKaroses thank you for your PR. Are you still interested in getting this in? If so, could you rebase this PR and resolve the conflicts.

vanniktech:
I'm closing this issue due to inactivity. If you have any further input on this PR, don't hesitate to reopen this PR or post a new one.

PR #59, opened 2015-05-21
make plugin work for multi-module android project with flatDirs

dbachelder:
some quick fixes to get things working for our project... not sure if you want these changes or not...

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,6 @@
-#Tue Nov 25 14:37:34 CET 2014
+#Thu May 21 06:48:31 PDT 2015
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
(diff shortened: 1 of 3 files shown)


ben-manes:
Thanks! What are the benefits between this approach and ignoring flatDir repositories? I can think of two possible ones.

First that when dependency was unresolved, due to ignoring the repository, there was no exceptional cause and thus no clear reason given in the log. That may have been partially due to the logging bug you fixed.

The other is that even though flatDir repositories lack metadata, the version may be inferred by the dependency name. This may make the dependency comparable against a version found in an external repository and provide meaningful update information. I'm not sure if this is what Gradle will actually due, though.

dbachelder:
I was still ending up with NPE's when run against our project. I think hiding the flatDir repos wasn't actually hiding the dependencies.

ben-manes:
Thanks! I wrote a unit test that reproduced the NPE, which your changes fix.

@jochenberger Please perform a release when you have the time.

Issue #62, opened 2015-06-16
Shows buildscript dependencies not project dependencies


shevek:
Since updating from gradle 2.2.1 and plugin version 0.4 to gradle 2.4 and plugin version 0.10.1, I seem to see the updates to the buildscript dependencies, not those of the project itself?

shevek:
Test case here: https://github.com/shevek/gradle-stdproject-plugin
Run `./gradlew dependencyUpdates`
Beware confusion: Build script deps are very similar to project deps. However, setting one to gradle-git 1.0.0 and the other to 1.2.0, I only see the updates for the buildscript, not the project.

shevek:
Note: On other projects where the deps do not overlap, it does work, so is this something to do with not reporting correctly if a buildscript dep is the same as a project dep, but of a newer version?

ben-manes:
I'll try to take a look at this tonight and convert your test case into a Spock unit test.

If I understand correctly, you have two dependencies with the same coordinates but different versions. One is up to date and the other isn't, and you expect the report to flag the outdated version? If that is correct then then I'm surprised it ever reported correctly because it globs all of the dependencies into a detached configuration for resolution. The normal dependency logic of taking the greatest version in a dependency set would apply.

shevek:
I do have two dependencies with the same coordinates and different versions, but they're also in different configurations: One's in the buildscript and one's in the compile scope. The test case should explain more, I think it's simple enough.

dependencyUpdates is working fine for other projects, so the question is "Why are buildscript dependencies leaking into project.compile dependencies"?

ben-manes:
Right, this is because the associated configurations are ignored. The plugin aggregates all of the repositories and dependencies across all configurations, adds them to a detached configuration, resolves that, and generates a report.

On the plus side this approach is very simple and covers the common usage. It leverages rather than replicates Gradle's dependency management. The flaws are that Gradle masks version skew (as it chooses the latest declared), does not use the configuration's resolution rules, and does not provide access to the full version history to apply custom filters when versions do follow the Maven/Ivy standard (e.g. "2.0-beta").

I think that instead of aggregating to a detached configuration, we could resolve every configuration and flatten that into a single report. The dependency cache should make this fast by only performing the external lookup once. Then resolution rules are honored and the duplicate coordinates are listed per version.

ben-manes:
[...]

Issue #61, opened 2015-06-16
Ignore snapshot?


henrik242:
Is there a way to get the plugin to ignore snapshot dependencies?

ben-manes:
Gradle supports revision tags as part of its resolution strategy when querying a dependency. The plugin should ignore snapshots by default and you can toggle the revision tag that is queries with. From the readme,

***

>The revision task property controls the resolution strategy of determining what constitutes the latest version of a dependency. The following strategies are supported:

>* release: selects the latest release
>* milestone: select the latest version being either a milestone or a release (default)
>* integration: selects the latest revision of the dependency module (such as SNAPSHOT)
The strategy can be specified either on the task or as a system property for ad hoc usage:

> `gradle dependencyUpdates -Drevision=release`

ben-manes:
Closing as no followup indicating this is a bug.

henrik242:
But I still see stuff like this when doing `gradle dependencyUpdates` (without any arguments):

Download http://mavenproxy.finntech.no/finntech-snapshot/org/apache/camel/camel-core/2.16-SNAPSHOT/camel-core-2.16-20150618.032519-96.pom
Download http://mavenproxy.finntech.no/finntech-snapshot/org/apache/camel/camel-parent/2.16-SNAPSHOT/camel-parent-2.16-20150618.033018-91.pom

Is that normal?  It can be quite a lot of them.

ben-manes:
Nope, that's a problem. The snapshots should be filtered by Gradle since it does the resolution for us.

Can you put a test sample together? I'd need to know the repositories, dependencies, and Gradle version to replicate.

henrik242:
Reproduce by running `gradle dependencyUpdates` with the following build.gradle:

    plugins {
      id "com.github.ben-manes.versions" version "0.9"
    }

    apply plugin: 'java'

    repositories {
      maven { url "https://repository.apache.org/content/repositories/snapshots/" }
      maven { url "https://repository.apache.org/content/repositories/releases/" }
    }

    dependencies {
      compile(
        "org.apache.camel:camel-core:2.13"
      )
    }

ben-manes:
Thanks! I'll try to take a look at this over the weekend.

ben-manes:
Oh, which Gradle version?

henrik242:
2.4

ben-manes:
```bash
$ gradle dU
:dependencyUpdates

------------------------------------------------------------
[...]

Issue #31, opened 2014-06-03
Support for revision=minor


henrik242:
I want to keep some revisions on their major releases, but want to upgrade if there is a new minor release.  One example is tomcat 8.0.x, which is still a beta product -- I want to stay at 7.x.x.

So maybe a -Drevision=minor could support this?

ben-manes:
Fixed in #63. You can now use component selection rules to customize the report.

Issue #30, opened 2014-05-03
It would be good to allow suppression of repositories.


tkruse:
Hi,my problem is that when using the version plugin and the grunt plugin like this:

    apply plugin: 'versions'
    apply plugin: 'grunt' // and grunt_tasks. implies node plugin
    buildscript {
    repositories {
        jcenter() // bintray jcenter faster than mavenCentral
        maven { url 'http://dl.bintray.com/robfletcher/gradle-plugins' }
    }
    dependencies {
        classpath 'com.moowork.gradle:gradle-grunt-plugin:0.5'
    }
    }
    node {
      version = '0.10.20'
      download = true
      workDir = file("${projectDir}/nodejs") // Set the work directory for unpacking node
    }

then the dependencyUpdates Task takes an eternity because the each dependency is tried 100 times against the nodejs repo:

     $ gradle dependencyUpdates --info
      Task has not declared any outputs.
    Resolving with repositories:
     - BintrayJCenter: http://jcenter.bintray.com/
    ...
     - ivy: http://nodejs.org/dist
    Resource missing. [HTTP GET: http://nodejs.org/dist/v0.11.13/ivy.xml]
    Resource missing. [HTTP HEAD: http://nodejs.org/dist/v0.11.13/powermock-classloading-xstream.jar]
    Resource missing. [HTTP GET: http://nodejs.org/dist/v0.11.12/ivy.xml]
    Resource missing. [HTTP HEAD: http://nodejs.org/dist/v0.11.12/powermock-classloading-xstream.jar]
    Resource missing. [HTTP GET: http://nodejs.org/dist/v0.11.11/ivy.xml]
    Resource missing. [HTTP HEAD: http://nodejs.org/dist/v0.11.11/powermock-classloading-xstream.jar]
    Resource missing. [HTTP GET: http://nodejs.org/dist/v0.11.10/ivy.xml]
    Resource missing. [HTTP HEAD: http://nodejs.org/dist/v0.11.10/powermock-classloading-xstream.jar]
    Resource missing. [HTTP GET: http://nodejs.org/dist/v0.11.9/ivy.xml]
    Resource missing. [HTTP HEAD: http://nodejs.org/dist/v0.11.9/powermock-classloading-xstream.jar]
    Resource missing. [HTTP GET: http://nodejs.org/dist/v0.11.8/ivy.xml]
[...]

PR #63, opened 2015-06-22
Major refactoring of versions plugin

ben-manes:
@jochenberger 

This is work in progress, but ready for discussion. The pending tasks are making all tests pass and removing boilerplate adaption code for passing the results into the reporter. These changes should help in addressing #28, #30, #61, and #62.

Most of the changes are cosmetic. The plugin was originally written in a scripting language approach (short but untyped). That coding style became awkward when converted to be type-safe, so it has been refactored to be more Java like.

Rather than using a detached configuration of all project dependencies, each configuration is copied and resolved independently. This should honor the configuration's resolution strategy (#28) and list each dependency version as an entry (#62). This also lets us decide whether to continue to resolve configurations with all repositories or only those associated with the given configuration. That would avoid suppressing problematic repositories (#30), such as the quirks caused by the plugin repository (#61).

Using resolution strategies or component rules may make it possible to add revision exclusions like "beta" and "rc" to the report (#25, #31). This could now be done in user code, and later evolved as a plugin configuration. #40 shows some promise and could be revived.

Performance should remain similar due to the Gradle cache making duplicate resolutions no-ops.


ben-manes:
All tests passing

ben-manes:
I think I'll defer refactoring the report to avoid the adapter code to another commit. The changes are now in a state ready to merge after discussion.

ben-manes:
I had some promising results [experimenting](https://discuss.gradle.org/t/excessive-downloads-during-resolution/10193/7) with component rules. Unfortunately it doesn't work correctly with `latest.${revision}`, but does fine with `+`. It may make sense to use `+` and then check the metadata status for backwards compatibility.

ben-manes:
@jochenberger 

Can you please merge this and release? I switched to `+` and it works beautifully with custom rules to reject a release candidate.

jochenberger:
Sorry, I haven't found the time to have a closer look, but I'll create a release right away.

ben-manes:
No problem, I got excited by the component rules working. So I pushed a release using my maven central account, which should sync with jcenter. My only concern is that the plugin portal may not be updated, but then users will realize when they run the task, right? ;-)

jochenberger:
I created the 0.11.1 release in JCenter, syncing from Maven Central does not seem to work reliably anyway. The new version should show up at https://plugins.gradle.org/plugin/com.github.ben-manes.versions soon.

ben-manes:
Thanks!

jochenberger:
Sorry, but I guess, we may need a 0.11.4. ;-)
I just got the advise to upgrade a library from 0.3.0 to 0.3.0. My project has a direct dependency on 0.1.2 and a transitive one on 0.3.0 and 0.3.0 is the latest version available, so it seems to be only a reporting issue.

ben-manes:
Does the report say `[0.1.2 -> 0.3.0]` or `[0.3.0 -> 0.3.0]`? If the former then would this be due to telling Gradle to not resolve a configuration transitively?

jochenberger:
It does indeed say `[0.3.0 -> 0.3.0]`. And both versions belong to the same configuration (compile). As far as I remember, previous versions of the plugin listed the library as up-to-date. But that might also be related to the new Gradle release candidate. I will have another look tomorrow, I'm AFK for today.

ben-manes:
Great, have a good evening. I've got to do the day job stuff anyways =)

jochenberger:
Alright, that's strange. It says `[0.3.0 -> 0.3.0]` in all the 0.11.x versions. 0.10.1 says that the library is up-to-date at 0.3.0 (which is the version that is actually used but not the one from the build file).

jochenberger:
Alright, it's related to my multi-project setup. The root project has a dependency on 0.1.2, a subproject depends on 0.3.0 (don't ask...).

ben-manes:
I wonder why this only shows up on 0.11.x and not previous releases.

The old approach grabbed all dependencies across all configurations. The new way resolves each configuration independently. So it should have been pulled in, perhaps inappropriately, by both.

ben-manes:
I'm going to try to fix the issues you found over the weekend. I'll try to find a few large Gradle projects to test on as well, to see if I can wrangle out a few other issues. Sorry for springing this on you out of the blue.

jochenberger:
No worries, it's fun to work on the plugin, there's always a lot to learn for me.
Maybe we should also try to increase the test coverage. I know that I've been quite lazy when it comes to writing tests...

ben-manes:
I fixed the `jacoco` dependency that you observed. It appears that Gradle hides some dependencies by not listing them in `configuration.dependencies` but they appear in the resolved set. That set is now filtered by the declared listing.

ben-manes:
Tested snapshot on
* Hibernate
* Spring
* Grails
* Forkhubs (android)
* cgeo (android)

The results look good and performance is comparable to the previous release.

jochenberger:
I just tested a snapshot, it still shows the jacoco and pmd libraries.

ben-manes:
hmm, I swore I had that fixed. Thanks, I'll take another look.

ben-manes:
Okay, I think I got it this time.

jochenberger:
Yes, looks good now, thanks!

Issue #104, opened 2016-06-07
[Minor bug] Depend on the OS when printing output.


zsavely:
I see an output:
`...\TestingProject\app\build/dependencyUpdates/report.txt` 
As you see, there are both backslashes and forward slashes. This probably affects nothing however copying from the console would be easier.

ben-manes:
Thanks for fixing this.

PR #106, opened 2016-06-07
Changed '/' to a system dependent separator.

zsavely:
Addresses https://github.com/ben-manes/gradle-versions-plugin/issues/104.


ben-manes:
Thanks!

PR #128, opened 2017-02-07
upgrade Gradle wrapper to 3.3

jochenberger:


jochenberger:
A more recent version is required for #129

ben-manes:
Its amusing they updated the shell script to correct "Windowz"

PR #129, opened 2017-02-07
add a test that ensures compatibility with different Gradle versions

jochenberger:
This only tests against Gradle 2.0 and 3.3. We can always add other versions later.


ben-manes:
Nice

PR #130, opened 2017-02-07
add Travis build

jochenberger:
@ben-manes, you probably have to enable the build from https://travis-ci.org/ben-manes/gradle-versions-plugin, I don't have the required permissions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,4 @@
+language: java
+cache:
+  directories:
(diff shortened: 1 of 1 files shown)


jochenberger:
Of course, if you prefer another CI provider, we can also use that one. I have only used Travis so far, so I can't advise on which one's best.

ben-manes:
I like Travis. Thanks!

PR #131, opened 2017-02-08
Support Gradle 3.4

ben-manes:
Work in progress. Doesn't handle resolution strategies properly, so may not be the correct approach.

@jochenberger

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-3.4-rc-2-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-3.3-bin.zip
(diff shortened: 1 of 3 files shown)


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-3.3-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-3.4-rc-2-bin.zip
```
jochenberger: You don't have to update the wrapper version to test against new Gradle versions. TestKit takes care of forking a build with the specified release.

jochenberger: In fact, I think we should not make this change in this PR, This just affects the Gradle version that is used to *build* the plugin. https://github.com/ben-manes/gradle-versions-plugin/pull/131/commits/ba4d411e955a4dff40a40c05c78585fed82fc515 adds 3.4-rc-2 to the versions to test against.

ben-manes: Sure, we can drop this. I was manually testing example build scripts

jochenberger:
That doesn't seem to be right, it also breaks Gradle 2.0 and 1.12 (and probably others that we don't test against).

PR #132, opened 2017-02-08
Fix for Gradle 1.12

jochenberger:
The plugin does not run with Gradle 1.12 anymore. It also fails for earlier versions, but let's take one step at a time.


jochenberger:
Probably caused by 122ba358bdd5d0a34dbb39b441f9aaf4cc31b597

PR #135, opened 2017-02-08
Deploy from travis

jochenberger:
Travis will automatically build tags and deploy them to the Bintray repository

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,3 +2,10 @@ language: java
 cache:
   directories:
   - $HOME/.gradle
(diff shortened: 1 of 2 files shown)


jochenberger:
@ben-manes That means that you can also create releases now. :-)

ben-manes:
Oh no. Now I need a new excuse to bug you :)

jochenberger:
Well, it doesn't mean that you *have to* create the releases. Also, we can still hope that it just doesn't work. ;-)

PR #143, opened 2017-02-21
Update README.md

grimreaper:
tested with 3.4 too

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ buildscript {
 }
 ```
 
-The current version is known to work with Gradle versions up to 3.3.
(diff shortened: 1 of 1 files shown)


ben-manes:
@jochenberger with your auto release, how do we merge changes without modifying a released jar? Does jcenter support snapshots?

jochenberger:
Travis will only deploy tagged versions to JCenter, so if you don't create a tag, nothing will change. :-)

ben-manes:
oh nifty. Thanks @jochenberger and @grimreaper

PR #178, opened 2017-09-19
Test against the most recent releases

jochenberger:
Make sure that the plug-in works with Gradle 4.1 and 4.2 (RC)


ben-manes:
Looks great as usual, thanks!

PR #209, opened 2018-03-10
Remove authorship since it is open source

jaredsburrows:
You can use `git log` or Github's "blame" to see past authors.


jaredsburrows:
@ben-manes What do you think of this one?

ben-manes:
I don't have a strong opinion on this, so either way is fine. I originally got into the habit to feel guilty for writing bad code (as my name attached) and later liked giving credit to authors. But none of that really justifies it if conventions have shifted with modern tools.

jaredsburrows:
If you have no problems, I will merge. I believe it harder to keep track and everyone can collaborate and see their name via `git log` or Github's blame.

PR #215, opened 2018-03-16
Add project urls from pom to dependency reports

### LINKED ISSUE #94: Output project url from pom
My usual steps when updating dependencies are:
1. check if there are new versions
2. go to project that was updated, to check changelog
3. update the dependency

This plugin makes step 1 and 3 are super easy and it would be nice if it could make step 2 almost as easy. Printing the project url of a changed dependency next to or below it would achieve that.

gabrielittner:
closes #94

### DIFF
--- a/src/test/resources/maven/backport-util-concurrent/backport-util-concurrent-java12/3.1/backport-util-concurrent-java12-3.1.pom
+++ b/src/test/resources/maven/backport-util-concurrent/backport-util-concurrent-java12/3.1/backport-util-concurrent-java12-3.1.pom
@@ -5,7 +5,6 @@
     <version>3.1</version>
     <packaging>jar</packaging>
     <name>Backport of JSR 166 for Java 1.2 and 1.3</name>
-    <url>http://backport-jsr166.sourceforge.net/</url>
     <description>Dawid Kurzyniec's backport of JSR 166 for Java 1.2 and 1.3</description>
     <licenses>
         <license>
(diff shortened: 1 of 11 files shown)


src/test/resources/maven/backport-util-concurrent/backport-util-concurrent-java12/3.1/backport-util-concurrent-java12-3.1.pom
```diff
@@ -5,7 +5,6 @@
     <version>3.1</version>
     <packaging>jar</packaging>
     <name>Backport of JSR 166 for Java 1.2 and 1.3</name>
-    <url>http://backport-jsr166.sourceforge.net/</url>
```
gabrielittner: Hope it's ok that I edited this and the following two to have test cases without adding more to the repo.

jaredsburrows: Why remove this? Just update the tests :P

ben-manes:
Can you please update the readme examples, too 😀

gabrielittner:
Done

vanniktech:
Thanks 👍

ben-manes:
I noticed that the artifact query is slow because Gradle doesn't cache it like dependency resolutions. I added a few optimizations that make it more bearable.

- Skip if offline is enabled (instead of ignored http failures)
- Use a lookup cache to resolve each project url once (if dependency is on multiple configurations)
- Lookup in parallel using GPars pool (but I think Gradle forces these to be synchronous)

This has a nice boost. We might need to add a flag if users complain about performance, but it already wasn't speedy.

Issue #76, opened 2015-08-23
[Feature Request] - Check Gradle Wrapper Version


jaredsburrows:
I use this plugin in all of my android apps and Gradle projects.

<pre>
$ tree gradle
gradle
`-- wrapper
    |-- gradle-wrapper.jar
    `-- gradle-wrapper.properties
</pre>

Can we add support to check for the latest Gradle version?

ben-manes:
It looks doable. The gradle distributions are not provided in a maven / ivy repository, so special handling of the [json routes](https://services.gradle.org/) would be required. The `release`, `milestone`, and `integration` revision types could be mapped to the distribution's `current`, `release candidate`, `nightly` revision types.

jaredsburrows:
@ben-manes Something like this: https://github.com/gradle/gradle/blob/ecb669edb1af1804e91eada92d123256ebb97fc8/gradle/wrapper.gradle#L17.

It works great:
```
Wrapper tasks
-------------
currentWrapper
nightlyWrapper
rcWrapper
```

Try copying to your build and running: `gradlew currentWrapper`.

ben-manes:
I don't think we'd want to generate the wrapper. That's built in, e.g.
`gradle wrapper --gradle-version 2.14`

I would think instead that we would want to display the version in the report (not sure if that should be in its own section or not). We'd fetch the version using those routes, query `gradle.gradleVersion`, and perhaps use the version comparator rather than a equality check (if we want to show release candidates).

I agree it should be fairly simple and a nice enhancement.

jaredsburrows:
@ben-manes Yeah. I am just showing how the Gradle team does it in order to update their local gradle. Since it can update it, we know the version, we can just print out the version.

jaredsburrows:
@ben-manes Something like this:
```
task("gradleUpdates") {
    group = "wrapper"
    ["nightly": "nightly",
     "rc"     : "release-candidate",
     "current": "current"].each {
        def label = it.value
        def version = new groovy.json.JsonSlurper().parseText(new URL("https://services.gradle.org/versions/${label}").text)
        if (version.empty) {
            throw new GradleException("Cannot update wrapper to '${label}' version as there is currently no version of that label")
        }
        println "${label} version: ${version.version} (downloadUrl: ${version.downloadUrl})"
    }
}
```

Command line:
```
$ gradlew gradleUpdates
```

Output:
```
$ gradle gradleUpdates
[...]

PR #222, opened 2018-04-14
Basic implementation of Gradle update checking

aaronols:
This pull request modifies the plugin to also query information from the public gradle releases api (https://services.gradle.org/versions/) as requested & discussed in https://github.com/ben-manes/gradle-versions-plugin/issues/76
It integrates the results of the gradle version checks into the existing `Result` object which makes it compatible with the automated JSON and XML exporters. The PlainTextReporter has been updated to output the Gradle Releases accordingly.

The magic sauce behind the entire thing is the class `org.gradle.util.GradleVersion`, which as it turns out implements `Comparable`, so I didnt have to deal with any Regex or string manipulation shenanigans to compare versions. And it also provides a static getter for the currently running version.

There are still a few things to do / to discuss before this should be considered to be merged:
- [x] Version number bump to 0.18.0 ok? -> no
- [x] consolidate Cache classes into UpdateChecker class
- [x] GroovyDoc comments
- [x] The implementation does not deal with errors / connection failures at the moment / check for `--offline` flag -> done. deals with errors. offline flag only affects gradle-internal artifact resolvers.
- [x] Fix Scopes / condense scope annotations
- [ ] Tests
- [x] Is it necessary to add new aliases to the XML Serializer? -> doesn't look like it
- [x] I used an enum with a constructor to hold the different release channels and their associated url information. This seems to break Tests on Gradle 1.x, which may be caused by the Groovy 1.8 runtime. Gradle 2.x executes tests without failure (maybe because it uses Groovy 2.3). When deciding how to proceed here, it should also be noted that Gradle 5.0 will remove support for running Gradle Builds older than 2.6. (verbatim output: `Support for builds using Gradle older than 2.6 was deprecated and will be removed in 5.0. You are currently using Gradle version 1.5. You should upgrade your Gradle build to use Gradle 2.6 or later.`) Other possible solution: check for gradle version before executing gradle update checking? -> removed tests running on Gradle 1.x
- [x] At the moment I have the gradle releases api url hardcoded, which means that the tests and each execution of the task will always try to connect to the gradle api for each time the task is executed (see backwards compatibility test). -> won't change
- [x] Put the download url to the releases in the report? (Personally, I don't download the distributions manually but update the wrapper task and execute a build so I wouldn't need this) -> not for now
- [x] remove nightly (only from TextReporter?) -> done
- [x] breadcrumb format for TextReporter (each arrow points towards a newer version than on the left side of the arrow) -> done
- [x] Update readme with information & examples -> done
- [x] License text for new files -> no because not practice in the rest of the project
- [x] squash commits in pull request -> done

Examples for how this looks in the output: (source formatting in Github text does not preserve spaces)
1. PlainTextReporter:
`------------------------------------------------------------`
`: Project Dependency Updates (report to plain text file)`
`------------------------------------------------------------`
` `
`Gradle release channels:`
`Running Gradle version: 4.6`
`- 'current' channel: [4.6] UP-TO-DATE`
`- 'release-candidate' channel: [4.6 -> 4.7-rc-2]`
`- 'nightly' channel: [4.6 -> 4.8-20180414000021+0000]`

2. JSON Reporter:
`{`
`    "current": { ... },`
`    "gradle": {`
`        "current": {`
`            "version": "4.6",`
`            "isUpdateAvailable": false`
`        },`
`        "nightly": {`
`            "version": "4.8-20180414000021+0000",`
`            "isUpdateAvailable": true`
`        },`
`        "releaseCandidate": {`
`            "version": "4.7-rc-2",`
`            "isUpdateAvailable": true`
`        },`
`        "running": {`
`            "version": "4.6",`
`            "isUpdateAvailable": false`
`        }`
`    },`
`    "exceeded": { ... },`
`    "outdated": { ... },`
`    "unresolved": { ... },`
`    "count": 4`
`}`

3. XML Reporter:
`<?xml version="1.0" encoding="UTF-8" standalone="yes"?>`
`<response>`
`  <count>4</count>`
`  <current> ... </current>`
`  <outdated> ... </outdated>`
`  <exceeded> ... </exceeded>`
`  <unresolved> ... </unresolved>`
`  <gradle>`
`    <running>`
`      <version>4.6</version>`
`      <isUpdateAvailable>false</isUpdateAvailable>`
`    </running>`
`    <current>`
`      <version>4.6</version>`
`      <isUpdateAvailable>false</isUpdateAvailable>`
`    </current>`
`    <releaseCandidate>`
`      <version>4.7-rc-2</version>`
`      <isUpdateAvailable>true</isUpdateAvailable>`
`    </releaseCandidate>`
`    <nightly>`
`      <version>4.8-20180414000021+0000</version>`
`      <isUpdateAvailable>true</isUpdateAvailable>`
`    </nightly>`
`  </gradle>`
`</response>`

### DIFF
--- a/src/main/groovy/com/github/benmanes/gradle/versions/reporter/PlainTextReporter.groovy
+++ b/src/main/groovy/com/github/benmanes/gradle/versions/reporter/PlainTextReporter.groovy
@@ -20,6 +20,7 @@ import com.github.benmanes.gradle.versions.reporter.result.DependencyLatest
 import com.github.benmanes.gradle.versions.reporter.result.DependencyOutdated
 import com.github.benmanes.gradle.versions.reporter.result.DependencyUnresolved
 import com.github.benmanes.gradle.versions.reporter.result.Result
+import com.github.benmanes.gradle.versions.updates.gradle.GradleReleaseChannel
 import groovy.transform.TupleConstructor
 import groovy.transform.TypeChecked
 
@@ -35,6 +36,8 @@ class PlainTextReporter extends AbstractReporter {
   /** Writes the report to the print stream. The stream is not automatically closed. */
   def write(printStream, Result result) {
     writeHeader(printStream)
+
+    writeGradleUpdates(printStream, result)
 
     if (result.count == 0) {
       printStream.println '\nNo dependencies found.'
@@ -56,6 +59,40 @@ class PlainTextReporter extends AbstractReporter {
       |------------------------------------------------------------
       |${project.path} Project Dependency Updates (report to plain text file)
       |------------------------------------------------------------""".stripMargin()
+  }
+
+  private def writeGradleUpdates(printStream, Result result) {
+    printStream.println('\nGradle updates:')
+
+    result.gradle.with {
+      // Log Gradle update checking failures.
+      if (current.isFailure) {
+        printStream.println("[ERROR] [release channel: ${GradleReleaseChannel.CURRENT.id}] " + current.reason)
+      }
+      if (releaseCandidate.isFailure) {
+        printStream.println("[ERROR] [release channel: ${GradleReleaseChannel.RELEASE_CANDIDATE.id}] " + releaseCandidate.reason)
+      }
+      if (nightly.isFailure) {
+        printStream.println("[ERROR] [release channel: ${GradleReleaseChannel.NIGHTLY.id}] " + nightly.reason)
+      }
+
+      // print Gradle updates in breadcrumb format
+      printStream.print(" - Gradle: [" + running.version)
+      boolean updatePrinted = false
+      if (current.isUpdateAvailable && current > running) {
+        updatePrinted = true
+        printStream.print(" -> " + current.version)
+      }
+      if (releaseCandidate.isUpdateAvailable && releaseCandidate > current) {
+        updatePrinted = true
+        printStream.print(" -> " + releaseCandidate.version)
+      }
+      // ignore nightly in textual output
+      if (!updatePrinted) {
+        printStream.print(" UP-TO-DATE")
+      }
+      printStream.println("]")
+    }
   }
 
   private def writeUpToDate(printStream, Result result) {
--- a/src/main/groovy/com/github/benmanes/gradle/versions/reporter/result/Result.groovy
+++ b/src/main/groovy/com/github/benmanes/gradle/versions/reporter/result/Result.groovy
@@ -1,5 +1,6 @@
 package com.github.benmanes.gradle.versions.reporter.result
 
+import com.github.benmanes.gradle.versions.updates.gradle.GradleUpdateResults
 import groovy.transform.TupleConstructor
 
 /**
@@ -28,4 +29,8 @@ class Result {
    * The unresolvable dependencies
    */
   DependenciesGroup<DependencyUnresolved> unresolved
+  /**
+   * Gradle release channels and respective update availability
+   */
+  GradleUpdateResults gradle
 }
--- a/src/main/groovy/com/github/benmanes/gradle/versions/updates/DependencyUpdates.groovy
+++ b/src/main/groovy/com/github/benmanes/gradle/versions/updates/DependencyUpdates.groovy
@@ -15,6 +15,7 @@
  */
 package com.github.benmanes.gradle.versions.updates
 
+import com.github.benmanes.gradle.versions.updates.gradle.GradleUpdateChecker
 import groovy.transform.TupleConstructor
 import groovy.transform.TypeChecked
 import org.gradle.api.Project
@@ -79,9 +80,12 @@ class DependencyUpdates {
     Map<Map<String, String>, String> downgradeVersions = toMap(versions.downgrade)
     Map<Map<String, String>, String> upgradeVersions = toMap(versions.upgrade)
 
+    // Check for Gradle updates.
+    GradleUpdateChecker gradleUpdateChecker = new GradleUpdateChecker()
+
     return new DependencyUpdatesReporter(project, revision, outputFormatter, outputDir, reportfileName,
       currentVersions, latestVersions, upToDateVersions, downgradeVersions, upgradeVersions,
-      unresolved, projectUrls)
+      unresolved, projectUrls, gradleUpdateChecker)
   }
 
   private static Map<Map<String, String>, String> toMap(Set<Coordinate> coordinates) {
--- a/src/main/groovy/com/github/benmanes/gradle/versions/updates/DependencyUpdatesReporter.groovy
+++ b/src/main/groovy/com/github/benmanes/gradle/versions/updates/DependencyUpdatesReporter.groovy
@@ -26,6 +26,9 @@ import com.github.benmanes.gradle.versions.reporter.result.DependencyOutdated
 import com.github.benmanes.gradle.versions.reporter.result.DependencyUnresolved
 import com.github.benmanes.gradle.versions.reporter.result.Result
 import com.github.benmanes.gradle.versions.reporter.result.VersionAvailable
+import com.github.benmanes.gradle.versions.updates.gradle.GradleUpdateChecker
+import com.github.benmanes.gradle.versions.updates.gradle.GradleUpdateResult
+import com.github.benmanes.gradle.versions.updates.gradle.GradleUpdateResults
 import groovy.transform.TupleConstructor
 import groovy.transform.TypeChecked
 import org.gradle.api.Project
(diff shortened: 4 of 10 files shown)


src/main/groovy/com/github/benmanes/gradle/versions/reporter/result/Result.groovy
```diff
@@ -28,4 +29,6 @@ class Result {
    * The unresolvable dependencies
    */
   DependenciesGroup<DependencyUnresolved> unresolved
+
+  GradleUpdateResults gradle
```
jaredsburrows: Groovy doc comment?

jaredsburrows:
Can you look into writing tests for this? Great work so far!

src/main/groovy/com/github/benmanes/gradle/versions/reporter/result/Result.groovy
aaronols: good point, I'll put doc comments on my code tomorrow

ben-manes:
I haven't had much time to review, but it looks good so far. A few small observations.

It feels like there may be unnecessary abstractions for the checking the versions. I would have expected maybe a checker, channel enum, and results. The http lookup, caching, etc, would be internal to the checker.

I agree errors and offline should be handled gracefully. If possible we should check for `--offline` too.

I think its fine to drop support for older Gradle versions. Alternatively, you could check the current version to disable this feature if too old (like we do in `Resolver`).

Nightly isn't too useful and most users won't look at the top (where up-to-date is listed). Perhaps the text report should only show the major/rc, at the bottom? That seems like the most important detail to call out, since text is for humans. It would also try to make it a single line, perhaps in a breadcrumb fashion often requested (current -> release -> rc)? The json/xml reports should stay as you implemented.

aaronols:
@jaredsburrows the update checker logic is executed implicitly by the `DifferentGradleVersionsSpec` test. Is this sufficient or do you want me to write tests explicitly for the new functionality?

@ben-manes I will move the cache class functionalities into the update checker.

Since Gradle 5.0 will remove support for running Builds with versions older than 2.6 I would like to remove the execution of the `DifferentGradleVersionsSpec` test for versions 1.5 and 1.12.

I agree that nightly is not very useful for users. Therefore i would remove it from the PlainTextReporter but keep it in the xml and json exports for programmatic evaluation.

I like the breadcrumb format. However it may not be appropriate to use for displaying information across multiple release channels. Normal format would be like you described `[4.6 -> 4.7 -> 4.8-rc1]` but there are a few corner cases I am unsure about:
- how do we deal with the case when there is a channel with no newer version than what the user already has installed. will the channel be displayed regardless `[4.6 -> 4.6 -> 4.7-rc1]` or will the channel be omitted `[4.6 -> 4.7-rc1]`? In case it is omitted, the user will have to interpret the breadcrumbs himself to determine which release channel the update is on
- how do we deal with the case when there is are multiple channels with updates newer than what the user has but not newer than each other? for example when a release candidate is available that is newer than what the user currently uses but older than the current release channel `4.6 -> 4.7 -> 4.7-rc3`?
a possible solution would be to check whether each of the breadcrumbs is newer than the previous breadcrumb. But then we are back at the first problem of which breadcrumbs to display.

ben-manes:
I thought maybe we'd omit channels that are older than the prior channel. Each arrow would point towards a newer version, if one exists.

```
[a] -> [b]
[a] -> [c]
[a] -> [b] -> [c]
[a] UP-TO-DATE
```

build.gradle
```diff
@@ -18,7 +18,7 @@ apply plugin: 'nexus'
 apply plugin: 'codenarc'
 
 group = 'com.github.ben-manes'
-version = '0.17.0'
+version = '0.18.0'
```
vanniktech: We don't need to make that change yet, do we?

src/main/groovy/com/github/benmanes/gradle/versions/updates/DependencyUpdates.groovy
```diff
@@ -78,9 +79,12 @@ class DependencyUpdates {
     Map<Map<String, String>, String> downgradeVersions = toMap(versions.downgrade)
     Map<Map<String, String>, String> upgradeVersions = toMap(versions.upgrade)
 
+    // check for gradle versions
```
vanniktech: nit: comments should be sentences https://nedbatchelder.com/blog/201401/comments_should_be_sentences.html

src/main/groovy/com/github/benmanes/gradle/versions/updates/DependencyUpdatesReporter.groovy
```diff
@@ -144,7 +150,21 @@ class DependencyUpdatesReporter {
...
       buildDependenciesGroup(outdated),
       buildDependenciesGroup(exceeded),
-      buildDependenciesGroup(unresolved)
+      buildDependenciesGroup(unresolved),
+      buildGradleUpdateResults()
+    )
+  }
+
+  /**
+   * Create a {@link  GradleUpdateResults} object from the information provided by the {@link GradleUpdateChecker}
```
vanniktech: nit: double space here.

src/main/groovy/com/github/benmanes/gradle/versions/updates/DependencyUpdatesReporter.groovy
```diff
@@ -144,7 +150,21 @@ class DependencyUpdatesReporter {
...
+   * @return filled out object instance
+   */
+  private GradleUpdateResults buildGradleUpdateResults() {
+    return new GradleUpdateResults(
+      running: new GradleUpdateResult(gradleUpdateChecker.runningGradleVersion, gradleUpdateChecker.runningGradleVersion),
```
vanniktech: What's the reason behind running here?

build.gradle
aaronols: No and i will revert this change. my reasoning was that i like to increment the version number before development.

src/main/groovy/com/github/benmanes/gradle/versions/updates/DependencyUpdatesReporter.groovy
aaronols: It is used to output the update path `4.6 -> 4.7` (with 4.6 being the 'running' version of gradle). It's also exported in the xml/json report and can then be programmatically evaluated.

aaronols:
I rebased the branch onto the current master, reverted the version number increment, consolidated the cache classes/logic into the GradleUpdateChecker and changed the comments as suggested.

ben-manes:
Sorry that I'm behind on everything, @aaronols. I hope to catch up a bit over the weekend, but I feel like I think that every week...

aaronols:
@ben-manes No worries :) I find myself having less time for spare time projects than i plan for, too. I have a modified implementation ready that deals with connection failure / no network / empty json when no release is available. After I look through the code again and verify it works as intended I'll push that in next few days.

aaronols:
I looked through the error handling logic again and pushed it.

aaronols:
In the latest commit i modified the PlainTextReporter to report gradle updates in breadcrumb format.
It will display [running -> current -> release-candidate]. Updates are skipped from display if they are not newer than the version on the left of the arrow.

The plugin output changes as follows:
`Gradle updates:`
` - Gradle: [4.6 -> 4.7 -> 4.8-rc-2]`

aaronols:
Updated the readme with infos about gradle update checking and squashed all commits.
I have implemented everything i wanted to and i would appreciate it if one of the collaborators would take another look on it.

ben-manes:
@aaronols If you're interested I'd be happy to give you the commit bit. I think anyone whose contributed can, if interested, be a collaborator (as there is really no harm to it). But like you said, its really hard to find the time for these projects so I wouldn't mean to impose that as making a commitment.

I'll try to take a look this during the week. It should be reviewed regardless of the above.

src/main/groovy/com/github/benmanes/gradle/versions/reporter/PlainTextReporter.groovy
```diff
@@ -36,6 +37,8 @@ class PlainTextReporter extends AbstractReporter {
   def write(printStream, Result result) {
     writeHeader(printStream)
 
+    writeGradleUpdates(printStream, result)
```
ben-manes: I think we should push this to the bottom, after dependencies instead of before. For projects with a lot of dependencies, this info will be lost as they watch the bottom half for things to update. So putting it last makes it standout, since upgrading the build tool is valuable.

ben-manes:
rebased and merged, thanks! 😄

PR #226, opened 2018-04-24
Custom filename for reports

### LINKED ISSUE #225: Renaming report file
Is there a way to rename the report file on generation time?

Poeschl:
This PR holds the option to specify the name of the generated report file.
It doesn't use gradle's build-in report features but for now does the job.

Its closes #225 und was inspired by #48 

I thought about a way to test the `reportfileName` option, but without success. If you have something in your mind about it or any remarks on the changes feel free to write.

### DIFF
--- a/src/main/groovy/com/github/benmanes/gradle/versions/updates/DependencyUpdatesReporter.groovy
+++ b/src/main/groovy/com/github/benmanes/gradle/versions/updates/DependencyUpdatesReporter.groovy
@@ -46,6 +46,8 @@ class DependencyUpdatesReporter {
   Object outputFormatter
   /** The outputDir for report. */
   String outputDir
+  /** The filename of the report file. */
+  String reportfileName
 
   /** The current versions of each dependency declared in the project(s). */
   Map<Map<String, String>, String> currentVersions
@@ -94,7 +96,7 @@ class DependencyUpdatesReporter {
   }
 
   def generateFileReport(Reporter reporter) {
-    String filename = outputDir + File.separator + reporter.getFileName()
+    String filename = outputDir + File.separator + reportfileName + '.' + reporter.getFileExtension()
     try {
       project.file(outputDir).mkdirs()
       File outputFile = project.file(filename)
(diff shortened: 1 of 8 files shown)


src/main/groovy/com/github/benmanes/gradle/versions/updates/DependencyUpdatesReporter.groovy
```diff
@@ -19,13 +19,7 @@ import com.github.benmanes.gradle.versions.reporter.JsonReporter
 import com.github.benmanes.gradle.versions.reporter.PlainTextReporter
 import com.github.benmanes.gradle.versions.reporter.Reporter
 import com.github.benmanes.gradle.versions.reporter.XmlReporter
-import com.github.benmanes.gradle.versions.reporter.result.DependenciesGroup
```
vanniktech: I don't know what the convention is here but I personally don't like star imports and since it has been changed I'd advise to revert this change.

src/main/groovy/com/github/benmanes/gradle/versions/updates/DependencyUpdatesTask.groovy
```diff
@@ -67,4 +70,7 @@ class DependencyUpdatesTask extends DefaultTask {
 
   /** Returns the outputDir destination. */
   String outputDirectory() { System.properties['outputDir'] ?: outputDir }
+
+  /** Returns the useProjectAsFilename boolean. */
```
vanniktech: Documentation is wrong here.

src/main/groovy/com/github/benmanes/gradle/versions/updates/DependencyUpdatesReporter.groovy
Poeschl: It seems that the company convention kicked in, will be reverted.

Poeschl:
Review integrated

vanniktech:
Thanks.

PR #233, opened 2018-05-30
Explain multi-project builds in README

skissane:
The README doesn't mention how dependencyUpdates works in multi-project builds.
I was running it separately in each subproject, when what I actually wanted was just to
run it in the root project – but I wasn't aware the later worked. So, add some text to the
README so others aren't confused like I was.


ben-manes:
Thanks!

PR #234, opened 2018-06-06
Issue#201: Resolve versions in parallel

rpau:
This patch resolves configurations in parallel using
gpars.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -48,7 +48,8 @@ dependencies {
   compile gradleApi()
   compile localGroovy()
 
-  compile 'com.thoughtworks.xstream:xstream:1.4.10'
+  compile 'com.thoughtworks.xstream:xstream:1.4.10',
+          'org.codehaus.gpars:gpars:1.2.1'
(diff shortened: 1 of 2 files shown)


ben-manes:
Thanks! Is there a reason you used gpars instead of Gradle’s worker api?
https://guides.gradle.org/using-the-worker-api/

rpau:
Yes, simplicity :) and because is the most widely used library for parallelizing collections in groovy.

ben-manes:
Maybe @bigdaz can chime in. Is there a pro/con of using Gradle’s worker api for a plugin versus gpars? Can workers be used in a backward compatible way, e.g. is injection optional and null check if the manager is absent?

@rpau I don't have a strong opinion, except concern that a Gradle facility might be better for some unknown implementation reason.

ben-manes:
rebased and merged, thanks!

ben-manes:
I tweaked your implementation a bit to parallelize all configurations across all projects, rather than process each project in lock-step.

PR #237, opened 2018-06-12
Make task configuration simpler for kotlin-dsl users and polish test setup

mkobit:
### Notes

* Gradle 4.8 minimum required for TestKit testing of the `kotlin-dsl` (see https://github.com/gradle/kotlin-dsl/issues/492 )


ben-manes:
Nice! Can you update the README to promote the method vs the instance field?

ben-manes:
I'd assume its just changing `dependencyUpdates.resolutionStrategy = {...}` to `dependencyUpdates.resolutionStrategy {...}`?

ben-manes:
Can you also look at the test failure?

mkobit:
> I'd assume its just changing dependencyUpdates.resolutionStrategy = {...} to dependencyUpdates.resolutionStrategy {...}?

Yup, it should just be that. Both ways should continue to work, but I didn't add any tests for the assignment way. Might not be a bad idea to have them, too.

I'll update the README and fix the test

ben-manes:
You should probably rebase, as the CI might be doing that automatically.

ben-manes:
verified that it breaks locally after a rebase. The reason is because of a bug on master, fixed in https://github.com/ben-manes/gradle-versions-plugin/commit/4c24d94b139c5709576a2cab43d96f531828409a, that swapped the group/name order.

mkobit:
Thanks for the fix @ben-manes - didn't realize yesterday that my `origin` remote also pointed to my fork :man_facepalming: - I've rebased off of master and updated the test and it looks good again locally. I also added a little bit to the README

ben-manes:
Looks great, thanks!

Just curious about Kotlin’s evaluation order. Can chaining be done like below or will the parser not evaluate the expression? I’ve seen languages on both sides of this.

```kotlin
val rejected = listOf("alpha", "beta", "rc", "cr", "m")
     .map { qualifier -> Regex("(?i).*[.-]$qualifier[.\\d-]*") }
     .any { it.matches(candidate.version) }

**Comment by mkobit** [comment]
Kotlin can parse that, and I think yours look cleaner so I pushed an update with it

**Comment by ben-manes** [comment]
I haven't ported any of my builds to the kotlin-dsl. Is there any way to do something like,

```kotlin
import com.github.benmanes.gradle.versions.updates.DependencyUpdatesTask
      
tasks {
  "dependencyUpdates"(DependencyUpdatesTask::class) {
  ...
}
```

That doesn't compile, but is more Java-esc by not qualifying classes inline. Just curious if there is an idiomatic way for that.

mkobit:
I would expect that to work if you bring the plugin into the classpath - here is an example where I do that https://github.com/mkobit/jenkins-scripts/blob/eee3d15badce442d4b00ea11888f3e4a4e4114e8/build.gradle.kts#L23

ben-manes:
weird. The test failed when I tried changing it. I thought it would look nicer in the README, but wasn't sure if it was safe. Just playing a bit before doing the release.

ben-manes:
Oh, my fault. The import has to be at the top of the file, in your `setup` section and I put it in the test's snippet.

PR #242, opened 2018-06-19
Add flag to disable gradle update check (fixes #241)

### LINKED ISSUE #241: Disable gradle version check for non-public wrapper URL
Short idea: fix gradle version check for non-official gradle distributions.

In our company we don't have access to the internet from gradle scripts. 

So, we use custom distribution URL, e.g. our gradle-wrapper.properties file is like this:

```
distributionBase=GRADLE_USER_HOME
distributionPath=wrapper/dists
zipStoreBase=GRADLE_USER_HOME
zipStorePath=wrapper/dists
distributionUrl=https\://<intranet-gradle-distribution-site>/org/gradle/binaries/gradle-4.8-all.zip
```

**The main issue:** plugin gets timeout error during the gradle version verification (and it works for a long time).

Possible solutions:
* Add ability to disable gradle updates check from task ```dependencyUpdates```
* Use distribution URL from file for next version lookup (e.g. for version 4.8 we can check 5.0, 4.9, 4.8.1, nothing more, because SemVer has three kinds)

ben-manes:
The task property `checkForGradleUpdate` is defaulted to `true`. When set to `false` no check is performed, the plain text report, does not include the statement, and the json/xml report includes an enable flag, no version, and the reason.

Due to URLConnection defaulting to infinite timeout that can cause a hung build, the maximum connection and read timeouts are set to 15 seconds as a failsafe.

@aaronols @imanushin

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,11 +50,11 @@ Displays a report of the project dependencies that are up-to-date, exceed the la
 have upgrades, or failed to be resolved. When a dependency cannot be resolved the exception is
 logged at the `info` level.
 
-Gradle updates are checked for on the `current`, `release-candidate` and `nightly` release channels.
-The plaintext report displays gradle updates as a separate category in breadcrumb style (excluding nightly builds).
-The xml and json reports include information about all three release channels, whether a release is considered an update
-with respect to the running (executing) gradle instance, whether an update check on on a release channel has failed, as
-well as a reason field explaining failures or missing information.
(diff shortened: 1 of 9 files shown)


imanushin:
@ben-manes , I suggest to patch documentation and put this option into the README.md file

imanushin:
@ben-manes , oh, sorry, I missed note. Thx

PR #273, opened 2018-11-12
272 Specify Gradle Release Channel

dje1990:
Specify Gradle Release Channel when checking for Gradle updates.

Don't tell me that there is a RC release for Gradle when I only want final release versions of Gradle, e.g. don't show me:
```
Gradle updates:
 - Gradle: [4.10.2 -> 5.0-rc-1]
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 bin
 build
+out
 .classpath
 .settings
 .project
--- a/src/main/groovy/com/github/benmanes/gradle/versions/reporter/AbstractReporter.groovy
+++ b/src/main/groovy/com/github/benmanes/gradle/versions/reporter/AbstractReporter.groovy
@@ -12,4 +12,6 @@ abstract class AbstractReporter implements Reporter {
   Project project
   /** The revision strategy evaluated with. */
   String revision
+  /** The gradle release channel to use for reporting. */
+  String gradleReleaseChannel
 }
--- a/src/main/groovy/com/github/benmanes/gradle/versions/reporter/PlainTextReporter.groovy
+++ b/src/main/groovy/com/github/benmanes/gradle/versions/reporter/PlainTextReporter.groovy
@@ -20,10 +20,10 @@ import com.github.benmanes.gradle.versions.reporter.result.DependencyLatest
 import com.github.benmanes.gradle.versions.reporter.result.DependencyOutdated
 import com.github.benmanes.gradle.versions.reporter.result.DependencyUnresolved
 import com.github.benmanes.gradle.versions.reporter.result.Result
-import com.github.benmanes.gradle.versions.updates.gradle.GradleReleaseChannel
 import groovy.transform.TupleConstructor
 import groovy.transform.TypeChecked
 
+import static com.github.benmanes.gradle.versions.updates.gradle.GradleReleaseChannel.*
 import static groovy.transform.TypeCheckingMode.SKIP
 
 /**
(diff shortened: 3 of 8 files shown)


dje1990:
To change the Gradle release channel use the new "gradleReleaseChannel" property, similar to the existing "revision" property
```
dependencyUpdates.gradleReleaseChannel="current"
dependencyUpdates.revision="release"
```

dje1990:
New outputs:

```
Gradle current updates:
 - Gradle: [4.10.2: UP-TO-DATE]
```

```
 - Gradle: [4.10.2 -> 5.0-rc-2]
```

```
Gradle nightly updates:
 - Gradle: [4.10.2 -> 5.1-20181112000041+0000]
```

Default is still release-candidate.

ben-manes:
Thanks for putting a PR with tests!

I do wonder about why you want this, though. An RC is good to be aware of as they often having breaking changes, so it gives you a chance to run and report issues early. Since its only a report, you don't have to take any action and the noise is pretty small. So I don't know if the option carries its weight, but maybe you have scenarios that I'm missing?

dje1990:
Please review and let me know of any suggestions for the implementation or it's tests. I'm interested in learning more about Groovy. (I would be more comfortable if the test coverage could be improved). The README.txt can also be updated to reflect this change.

dje1990:
Just a personal preference. I'm not going to download each RC version of Gradle locally, and there's not RC version of Gradle available in the official Gradle Docker image at https://hub.docker.com/_/gradle/ so I'd rather it not report in my build logs, one less thing to think about.

ben-manes:
A `resolutionStrategy` is for handling dependency filtering. I think that's still be best approach because Maven's metadata is too limited and there are a variety of versioning schemes in use.

This change only applies to the text/console report, not to the json/xml versions. That differs from all of the other features. Should we begin having a report block for type specific customizations? I don't particularly like mixing them together at the top, as it could be more confusing long term.

corecanarias:
Hi, I'm interested on this changes as well. To give a bit of extra context on why, I came to this plugin from https://github.com/patrikerdes/gradle-use-latest-versions-plugin and after that I started to take a look at https://renovatebot.com/ that sends pull requests with recommended upgrades for your dependencies. 

Renovate didn't support any java project, so I added the support for gradle using gradle-versions-plugin as a way to get a list of dependencies to update that then, renovate will use to create the pull requests.

But now, we are seeing that we need some additional features from gradle-versions-plugin to give a good user experience to renovate users. 

I see that we could use the gradle resolutionStrategy to solve that. But for us makes more sense having that logic in renovate shared among all the different projects, than having to add it to every project. Specially because the expectations is that renovate only support a specific set of use cases, normally only want to receive pull requests from specific updates, like only minor updates, no rc or snapshotsso. It shouldn't be too far away from what this pr already provides if it applies to the json/xml reports which is what we use in renovate.

dje1990:
Hi @ben-manes @corecanarias 

Regarding JSON/XML reports,

From JSON reporter:
```
  def write(printStream, Result result) {
    printStream.println new JsonBuilder(result).toPrettyString().stripMargin()
  }
```
From XML reporter:
```
  @Override
  def write(printStream, Result result) {

    XStream xstream = new XStream()
    xstream.with {
      alias('response', Result)
      alias('available', VersionAvailable)
      alias('exceededDependency', DependencyLatest)
      alias('outdatedDependency', DependencyOutdated)
      alias('unresolvedDependency', DependencyUnresolved)
      alias('dependency', Dependency)
      alias('group', DependenciesGroup)
    }

    printStream.println '<?xml version="1.0" encoding="UTF-8" standalone="yes"?>'
    printStream.println xstream.toXML(result).stripMargin()
  }
```
The `Result `object contains property `GradleUpdateResults gradle` 
`GradleUpdateResults` contains 
```
@TupleConstructor
class GradleUpdateResults {
  boolean enabled
  GradleUpdateResult running
  GradleUpdateResult current
  GradleUpdateResult releaseCandidate
  GradleUpdateResult nightly
}
```
And each `GradleUpdateResult` contains 
```
  GradleUpdateResult(boolean enabled, GradleUpdateChecker.ReleaseStatus.Available running, GradleUpdateChecker.ReleaseStatus release) {
    if (!enabled) {
      this.version = ""
      this.isUpdateAvailable = false
      this.isFailure = false
      this.reason = "update check disabled"
    } ... 
```

So it would be possible to go to the JSON/XML result inspect 
`gradle.nightly.reason == "update check disabled"`
`gradle.releaseCandidate.reason == "update check disabled"`
`gradle.current.reason == "update check disabled"`
to deduce what was the value of `dependencyUpdates.gradleReleaseChannel`

As a more simple approach, the `gradleReleaseChannel` could be added explicitly to the `Result` object but this would require further analysis as it could be a breaking change for some users.

Please let me know if you have any questions.
Thanks.

ben-manes:
Thanks @dje1990!

To clarify and summarize the toggles,
 - `revisision` is an Apache Ivy feature, back when Gradle was built on top of it, and was the only option we had in those early days. There is no Maven pom metadata in use that corresponds, so it only toggles between including or excluding `-SNAPSHOT` versions.
 - `resolutionStrategy` is the Gradle way for filtering dependencies. There are a variety of versioning styles (Maven, semver, custom) used in the Java world and it is very inconsistent. There is no metadata to discern from, so one has to resort to regex and conventions.
 - `gradleReleaseChannel` is for the Gradle distribution itself, which is oddly a different mechanism than dependencies. This feature from @dje1990 lets us report only on the channels of interest.

Unfortunately there is not a clear enough set of rules for a "release" version to set globally. Whatever we decided on would have to be revised regularly to handle odd ball dependencies. Some, for example, use even (release) / odd (beta) versions. Instead we have to leave it up to users to customize the reports as they see fit.

PR #280, opened 2018-12-21
Use task configuration avoidance.

runningcode:


.gitignore
```diff
@@ -9,3 +9,5 @@ out
 *.iml
 *.ipr
 *.iws
+
+local.properties
```
jaredsburrows: Was this added/autogenerated by your setup locally? Usually this is auto-generated by the Android Gradle Plugin.

runningcode: Yeah, the `local.properties` file is auto-generated by Android Studio. I added it to the `.gitignore` to prevent myself and others from accidentally committing it. I think it could be nice to keep here, but will happily revert.

ben-manes:
Thanks!

PR #291, opened 2019-02-14
Configure gradle build scan plugin

aalmiray:
This small PR configures the `build-scan` Gradle plugin (https://guides.gradle.org/creating-build-scans/) which yields further insight into the build, allowing the team to make informed decisions on where and when the build may be tweaked to get faster, more performant builds.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,7 +9,13 @@ buildscript {
 }
 
 plugins {
+  id "com.gradle.build-scan" version "1.16"
   id "com.jfrog.bintray" version "1.7.3"
+}
+
+buildScan {
+  termsOfServiceUrl = 'https://gradle.com/terms-of-service'
+  termsOfServiceAgree = 'yes'
 }
 
 apply plugin: 'java-gradle-plugin'
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ language: java
 cache:
   directories:
   - $HOME/.gradle
+script:
+  - ./gradlew build --scan
 deploy:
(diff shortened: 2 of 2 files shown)


build.gradle
```diff
@@ -9,9 +9,15 @@ buildscript {
 }
 
 plugins {
+  id "com.gradle.build-scan" version "1.16"
```
jaredsburrows: Can you use the 2.0+ version? Requires Gradle 5.0.
See: https://github.com/jaredsburrows/android-gif-example/blob/master/gradle/scan.gradle.kts

aalmiray: The current build sets `4.10.1` in the wrapper, that's why the chosen version of `build-scan` is `1.16'. If the build migrates to `5.x` then yes, by all means let's upgrade the `build-scan` plugin :smile:

ben-manes: There is no reason not to update, I don't think there should be a compatibility issue to do so. Can you give it a quick try?

aalmiray: There's one potential issue though. Gradle `5.x` includes a more recent version of Groovy and that may cause a compatibility issue. If upgraded then the next version of this plugin would force consumers to use Gradle `5.x`. Is that desirable?

ben-manes: That's a great point. I'm fine either way since the older versions are available, but would lean towards your mindset of holding back a bit to allow 5.x to spread into less actively maintained projects. What do you think @jaredsburrows?

aalmiray:
Thanks! :smile:

PR #298, opened 2019-03-19
Update example filters for OpenJFX and JAXB

mwkroening:
Changes are required to exclude release candidates from OpenJFX and JAXB.

JAXB example: org.glassfish.jaxb:jaxb-bom:2.4.0-b180830.0438
OpenJFX example: org.openjfx:javafx:13-ea+2


ben-manes:
thanks!

PR #327, opened 2019-08-27
Feature/current version in component selection

ghus-raba:
After facing similar issue as in https://github.com/ben-manes/gradle-versions-plugin/issues/313 and https://github.com/ben-manes/gradle-versions-plugin/issues/290, I had a look at how we could get current version to resolution strategy. 

I went for extending default gradle befavior by decorating the resolution strategy classes. Please let me know if the approach would be acceptable. As for code style, I am not much at home in groovy, so any improvements are welcome.

PR #330, opened 2019-08-29
Check if original dependency exists in component selection

ghus-raba:
Fixes https://github.com/ben-manes/gradle-versions-plugin/pull/327#issuecomment-526015459


ben-manes:
This seems pretty solid now, thanks for the quick fix :)

I'll try to debug the TravisCI issues. Their change to a new default distribution, Xenial, correlates to when the errors started to occur. It may be as easy as adding `dist: trusty`. Otherwise it might take a little time to work through the instability as we release by creating a new tag which TravisCI will publish.

PR #332, opened 2019-08-30
Remove unnecessary/incorrect escaping in examples

chadlwilson:
Extra escaping was added in c0b850601d74fd961762e6c8a80be980608714da and appears to break the matching; because $ is being treated as literal. Additionally, escaping the `\d` isn't necessary in a Groovy regex. 

Otherwise the new feature from #327 looks great - thanks!


chadlwilson:
No problem!

PR #333, opened 2019-08-31
Backwards compatibility of direct closure assignment (fixes #331)

### LINKED ISSUE #331: Breaking change in 0.23.0?
I've just picked up the new version of your library and we now get the following error when running `useLatestVersions`:

Could not set unknown property `'resolutionStrategy' for task ':dependencyUpdates' of type com.github.benmanes.gradle.versions.updates.DependencyUpdatesTask.`

If we change our gradle config from `dependencyUpdates.resolutionStrategy` to 
`dependencyUpdates.resolutionStrategyAction` things work as before.

Just wondering if this was an intentional change or not before we roll this fix out across our projects?

ben-manes:


src/main/groovy/com/github/benmanes/gradle/versions/updates/DependencyUpdatesTask.groovy
```diff
@@ -53,7 +54,8 @@ class DependencyUpdatesTask extends DefaultTask {
   boolean checkForGradleUpdate = true
 
   Object outputFormatter = 'plain'
-  Action<? super ResolutionStrategyWithCurrent> resolutionStrategyAction = null
+  Closure resolutionStrategy = null;
```
ghus-raba: If it was unintentional, then I would not worry about supporting this till the end of times. I get that you don't want to suddenly break the build for some people. In the future, it could be marked as deprecated with explanation + warning in the readme.

Otherwise LGTM

ben-manes: This was hidden in the guts of Gradle, but now shows up as a deprecated warning and the message visible when using `--warning-mode=all`.

Issue #325, opened 2019-08-22
Simplified resolutionStrategy


jmfayard:
**Update: Released in v0.25.0**

Hello Ben,

I propose to add a declarative filter for filtering version

```kotlin
rejectVersionIf {
    isNonStable(candidate.version) && !isNonStable(currentVersion)
}
```

instead of 

```kotlin
  resolutionStrategy {
    componentSelection {
      all {
        if (isNonStable(candidate.version) && !isNonStable(currentVersion)) {
          reject("Release candidate")
        }
      }
    }
  }
```

Also I noticed that people are *much* more creative on how they name their unstable versions than on how they name their unstable versions. 

So the `README`  should recommand something like:

```kotlin
fun isNonStable(version: String): Boolean {
  val stableKeyword = listOf("RELEASE", "FINAL", "GA").any { version.toUpperCase().contains(it) }
  val regex = "^[0-9,.v-]+$".toRegex()
  val isStable = stableKeyword || regex.matches(version)
  return isStable.not()
}
```

ben-manes:
Thanks, it’s very nice. Unfortunately it has to be empty by default. We don’t want to change behavior so opt-in rather than your opt-out.

jmfayard:
Yes, that was my intention as well. Leave it empty by default, but in the README document first this solution

chadlwilson:
This looks quite nice - quite a bit cleaner than the current approach.

jmfayard:
I was thinking how to best integrate it with the newest feature.

We could have a syntax

```
rejectVersionContaining(...)
```

which would be a shortcut for

```
rejectVersionIf { candidate ->
  candidate.version.matches(Regex("(?i).*[.-]\$qualifier[.\\d-+]*"))
}
```

which would be a shortcut for

resolutionStrategy {
    componentSelection {
      all {
        fun isNonStable(version: String) = listOf("alpha", "beta",
"rc", "cr", "m", "preview", "b", "ea").any { qualifier ->
          version.matches(Regex("(?i).*[.-]\$qualifier[.\\d-+]*"))
        }
        if (isNonStable(candidate.version) && !isNonStable(currentVersion)) {
          reject("Release candidate")
        }
      }
    }
  }

jmfayard:
i was thinking how to integrate it with the newest feature

we could have 

```kotlin

rejectVersionContaining ("alpha","beta",...)

```

which would be a shorter version of

```kotlin

rejectVersionIf { candidate ->

[...]

PR #337, opened 2019-09-06
WIP: rejectVersionIf { }  

jmfayard:
I had a first try  at implementing #325 

```kotlin
rejectVersionIf { current: ComponentSelectionWithCurrent ->
        isNonStable(current.candidate.version) || !isNonStable(current.currentVersion)
}
```

It would be better if we can change the syntax to this but I don't know how to do it:

```kotlin
rejectVersionIf { candidate: ModuleComponentIdentifier ->
        isNonStable(candidate.version) || !isNonStable(currentVersion)
}
```

`isNonStable` being defined as this:

```kotlin
fun isNonStable(version: String): Boolean {
  val stableKeyword = listOf("RELEASE", "FINAL", "GA").any { version.toUpperCase().contains(it) }
  val regex = "^[0-9,.v-]+$".toRegex()
  val isStable = stableKeyword || regex.matches(version)
  return isStable.not()
}
```

Since my Groovy skills are closed to zero, I tried to upgrade Gradle so that at least that part is familiar for me. 

Because of the Gradle update, one test is not working anymore https://scans.gradle.com/s/unwddfvrt5xak

```
    > Could not find method componentSelection() for arguments [build_6aatid6bxisq96bm8bf31j5zh$_run_closure3$_closure5@243ebea0] on root project 'junit8907024250739898265' of type org.gradle.api.Project.

dependencyUpdates.resolutionStrategy = componentSelection {
        all(new Rule())
   }
}
```

I have also run `./gradlew publish` and then use the artifact in my own plugin

https://github.com/jmfayard/buildSrcVersions/pull/62/files

One part of this is also wrong:

`Cannot change dependencies of configuration ':classpath' after it has been resolved.`

https://gist.github.com/jmfayard/f8455c2cdc658035e8d542ecf0f841ac

### DIFF
--- a/src/main/groovy/com/github/benmanes/gradle/versions/updates/DependencyUpdates.groovy
+++ b/src/main/groovy/com/github/benmanes/gradle/versions/updates/DependencyUpdates.groovy
@@ -22,6 +22,7 @@ import groovy.transform.TypeChecked
 import org.gradle.api.Action
 import org.gradle.api.Project
 import org.gradle.api.artifacts.Configuration
+import org.gradle.api.artifacts.ConfigurationContainer
 import org.gradle.api.artifacts.UnresolvedDependency
 
 /**
@@ -48,8 +49,9 @@ class DependencyUpdates {
 
   /** Evaluates the dependencies and returns a reporter. */
   DependencyUpdatesReporter run() {
-    Map<Project, Set<Configuration>> projectConfigs = project.allprojects.collectEntries { proj ->
-      [proj, proj.configurations.plus(proj.buildscript.configurations) as Set]
+    Map<Project, ConfigurationContainer> projectConfigs = project.allprojects.collectEntries { proj ->
+      proj.configurations.addAll(proj.buildscript.configurations)
+      [proj, proj.configurations]
     }
     Set<DependencyStatus> status = resolveProjects(projectConfigs)
 
@@ -62,7 +64,7 @@ class DependencyUpdates {
     return createReporter(versions, unresolved, projectUrls)
   }
 
-  private Set<DependencyStatus> resolveProjects(Map<Project, Set<Configuration>> projectConfigs) {
+  private Set<DependencyStatus> resolveProjects(Map<Project, ConfigurationContainer> projectConfigs) {
     projectConfigs.keySet().collect { proj ->
       Set<Configuration> configurations = projectConfigs.get(proj)
       Resolver resolver = new Resolver(proj, resolutionStrategy)
--- a/src/main/groovy/com/github/benmanes/gradle/versions/updates/DependencyUpdatesTask.groovy
+++ b/src/main/groovy/com/github/benmanes/gradle/versions/updates/DependencyUpdatesTask.groovy
@@ -15,10 +15,14 @@
  */
 package com.github.benmanes.gradle.versions.updates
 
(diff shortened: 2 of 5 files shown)


ben-manes:
If you use `gradle install` it will install into the local maven repository. Then you can use `mavenLocal()` in your repository to reference it, if you bump the version up. That's how I do ad hoc testing in the `examples` build, to see it in action.

I think Gradle changed the version of Groovy which is backwards incompatible, so we stayed on an old version which is forward compatible. So there could be some weird interaction there.

If you drop all your Gradle changes and use `mavenLocal`, does all of this work right?

src/main/groovy/com/github/benmanes/gradle/versions/updates/DependencyUpdatesTask.groovy
```diff
@@ -91,6 +98,18 @@ class DependencyUpdatesTask extends DefaultTask {
     this.resolutionStrategy = null
   }
 
+  void rejectVersionIf(final Transformer<Boolean, ComponentSelectionWithCurrent> filter) {
+    resolutionStrategy { ResolutionStrategyWithCurrent strategy ->
```
ben-manes: Sounds like you can now drop the these now that the delegation is fixed. Does it work like this?

```gradle
resolutionStrategy {
  strategy.componentSelection {
    all {
      if (filter.transform(current)) {
        current.reject("Rejected")
      }
    }
  }
}
```

ben-manes: oh, maybe not without losing the typesafe checking

src/main/groovy/com/github/benmanes/gradle/versions/updates/DependencyUpdatesTask.groovy
```diff
@@ -54,11 +58,14 @@ class DependencyUpdatesTask extends DefaultTask {
   @Input
   boolean checkForGradleUpdate = true
 
+  @Input @Optional
   Object outputFormatter = 'plain'
 
+  @Input @Optional
```
ghus-raba: This is causing the fail in the test, because the closure is executed without correct arguments and delegates during validation. Something similar seems to have happened in https://github.com/ben-manes/gradle-versions-plugin/commit/3cce551e9ced67e38dc99022018f209fd103bea1, where they removed the annotation from the `outputFormatter` field.

src/main/groovy/com/github/benmanes/gradle/versions/updates/DependencyUpdates.groovy
```diff
@@ -48,8 +49,9 @@ class DependencyUpdates {
 
   /** Evaluates the dependencies and returns a reporter. */
   DependencyUpdatesReporter run() {
-    Map<Project, Set<Configuration>> projectConfigs = project.allprojects.collectEntries { proj ->
-      [proj, proj.configurations.plus(proj.buildscript.configurations) as Set]
+    Map<Project, ConfigurationContainer> projectConfigs = project.allprojects.collectEntries { proj ->
+      proj.configurations.addAll(proj.buildscript.configurations)
```
ghus-raba: I think this could be the cause of `Cannot change dependencies of configuration ':classpath' after it has been resolved.` Try:
```groovy
def configurations = []
configurations.addAll(proj.configurations)
configurations.addAll(proj.buildscript.configurations)
[proj, configurations as Set]
```

Issue #344, opened 2019-09-26
New resolution  strategy doesn't work


timja:
Project used is:
https://github.com/hmcts/draft-store (latest commit at time of writing is 565c2f0d0e95bd36f161282aef3a72d0b41f05e7)
but I've heard of the same issues on other applications

```
$ g --version

------------------------------------------------------------
Gradle 5.6.2
------------------------------------------------------------

Build time:   2019-09-05 16:13:54 UTC
Revision:     55a5e53d855db8fc7b0e494412fc624051a8e781

Kotlin:       1.3.41
Groovy:       2.5.4
Ant:          Apache Ant(TM) version 1.9.14 compiled on March 12 2019
JVM:          1.8.0_222 (AdoptOpenJDK 25.222-b10)
OS:           Mac OS X 10.14.6 x86_64
```

```
dependencyUpdates {
  resolutionStrategy {
    componentSelection {
      all {
        if (isNonStable(candidate.version) && !isNonStable(currentVersion)) {
          reject('Release candidate')
        }
      }
    }
  }
}
```

```
Caused by: org.gradle.internal.metaobject.AbstractDynamicObject$CustomMessageMissingMethodException: Could not find method isNonStable() for arguments [2.9.10] on object of type com.github.benmanes.gradle.versions.updates.resolutionstrategy.ResolutionStrategyWithCurrent.
        at org.gradle.internal.metaobject.AbstractDynamicObject.methodMissingException(AbstractDynamicObject.java:182)
        at org.gradle.internal.metaobject.ConfigureDelegate.invokeMethod(ConfigureDelegate.java:86)
        at org.codehaus.groovy.runtime.metaclass.ClosureMetaClass.invokeOnDelegationObjects(ClosureMetaClass.java:398)
        at org.codehaus.groovy.runtime.metaclass.ClosureMetaClass.invokeMethod(ClosureMetaClass.java:338)
        at groovy.lang.MetaClassImpl.invokeMethod(MetaClassImpl.java:1041)
        at groovy.lang.MetaClassImpl.invokeMethod(MetaClassImpl.java:821)
        at groovy.lang.GroovyObjectSupport.invokeMethod(GroovyObjectSupport.java:44)
        at org.codehaus.groovy.runtime.metaclass.ClosureMetaClass.invokeOnDelegationObjects(ClosureMetaClass.java:398)
[...]

PR #346, opened 2019-09-28
Fix resolution strategy for Groovy.

johnjohndoe:
+ Before no versions were output.
+ This fix is based on the discussion in https://github.com/ben-manes/gradle-versions-plugin/issues/344.

---

@timja @jmfayard @bric3 Please review this fix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,19 +130,19 @@ You can either use the simplified syntax `rejectVersionIf { ... }` or configure 
 dependencyUpdates {
   // Example 1: reject all non stable versions
   rejectVersionIf {
-    isNonStable(candidate.version)
+    isNonStable(it.candidate.version)
   }
 
   // Example 2: disallow release candidates as upgradable versions from stable versions
   rejectVersionIf {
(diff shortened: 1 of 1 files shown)


jmfayard:
Lgtm

johnjohndoe:
@timja I agree. It is not obvious from the splitted code snippets in the README that the `isNonStable` variable can be defined in the `dependencyUpdates` block. However, it can. I tested both options. If `isNonStable` is defined outside the `dependencyUpdates` block then it works as described. If `isNonStable`  is defined inside the `dependencyUpdates` block then `it` has to be replaced with `qualifier` to avoid conflicts with another `it`.
I personally prefer to **limit the scope** the variable inside the `dependencyUpdates` block. 
For this pull request I decided that this should be discussed separately.

ben-manes:
Thanks!

bric3:
I would have the example split. But this looks good to me.

ben-manes:
Happy to accept a PR if you'd like improvements :)

bric3:
I know, I know 😉

PR #349, opened 2019-10-05
Escape . and - correctly in `isNonStable()` RegEx

floscher:
Since the `.` matches every character in a RegEx, when you don't escape it every version is considered stable.
Also I added an optional `-r` to the end, because for example `jgit` uses this in version numbers of stable releases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ no agreed standard on this, but this is a good starting point:
 ```groovy
 def isNonStable = { String version ->
   def stableKeyword = ['RELEASE', 'FINAL', 'GA'].any { it -> version.toUpperCase().contains(it) }
-  def regex = /^[0-9,.v-]+$/
+  def regex = /^[0-9,\\.v\\-]+(-r)?$/
   return !stableKeyword && !(version ==~ regex)
 }
 ```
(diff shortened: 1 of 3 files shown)


ben-manes:
thanks!

jmfayard:
> Since the . matches every character in a RegEx, when you don't escape it every version is considered stable.

Are you sure @floscher? I am surprised by this claim. 
[I had tests for `isNonStable`](https://github.com/jmfayard/buildSrcVersions/blob/e117dc0898aadbc1fce1029a87c48d5aa0ce269c/plugin/src/test/kotlin/de/fayard/UnitTests.kt#L62-L72) and the regex tester also indicates that `.` inside bracket does not match everything
 https://www.regextester.com/?fam=111778

floscher:
@jmfayard I now did some more testing and it turns out, `.` indeed doesn't need escaping inside the brackets, that claim is indeed wrong. Thanks for noticing that!
The `-` is a special character inside brackets, but it seems you don't necessarily have to escape it if it is directly before the closing bracket.
I had a slightly different order inside the bracket when I noticed my `isNotStable()` method was behaving wrong: `^[0-9.,-v]+$`. It then also matched versions like `1.2.3-alpha`. So I escaped both `.` and `-` and it worked afterwards, so I thought these were the causes and then opened this pull request.

But as it now turns out, in your case there was no escaping necessary at all (escaping has no negative effects, but is not needed).

PR #351, opened 2019-10-08
Resolve dependency constraints.

### LINKED ISSUE #350: Resolve versions in a java-platfom project
I usually manage dependency versions in a repo with many Gradle projects by having one `java-platform` project that uses `dependencies.constraints` to force versions in other projects. Currently, I usually run `dependencyUpdates` on the entire repo, meaning for all the subprojects, but it would be convenient to be able to just run `:project-bom:dependencyUpdates` to resolve updates to the constraint dependencies. Is this a feature that's appropriate for this plugin?

anuraaga:
Dependency constraints can be used to set versions, and in particular can be used with the `java-platform` plugin to define all constraints in a single location. It would be convenient if these constraints were resolved by the plugin since then instead of resolving the dependencies of all projects in a monorepo controlled by a platform, just need to resolve a single project for better performance and simpler output.

I am wondering whether this feature should be restricted to projects with the java-platform plugin. Let me know if you have thoughts.

Reference
- https://docs.gradle.org/current/userguide/managing_transitive_dependencies.html#sec:dependency_constraints
- https://docs.gradle.org/current/userguide/java_platform_plugin.html

Fixes #350

### DIFF
--- a/src/main/groovy/com/github/benmanes/gradle/versions/updates/Resolver.groovy
+++ b/src/main/groovy/com/github/benmanes/gradle/versions/updates/Resolver.groovy
@@ -25,6 +25,7 @@ import org.gradle.api.artifacts.ComponentMetadata
 import org.gradle.api.artifacts.ComponentSelection
 import org.gradle.api.artifacts.Configuration
 import org.gradle.api.artifacts.Dependency
+import org.gradle.api.artifacts.DependencyConstraint
 import org.gradle.api.artifacts.ExternalDependency
 import org.gradle.api.artifacts.LenientConfiguration
 import org.gradle.api.artifacts.ModuleVersionIdentifier
@@ -117,6 +118,15 @@ class Resolver {
       createQueryDependency(dependency, revision)
     }
 
+    // Common use case for dependency constraints is a java-platform BOM project.
+    try {
+      configuration.dependencyConstraints.each { dependency ->
+        latest.add(createQueryDependency(dependency, revision))
+      }
+    } catch (MissingPropertyException e) {
+      // Skip if constraints not supported
+    }
+
     Configuration copy = configuration.copyRecursive().setTransitive(false)
     // https://github.com/ben-manes/gradle-versions-plugin/issues/127
     if (copy.metaClass.respondsTo(copy, "setCanBeResolved", Boolean)) {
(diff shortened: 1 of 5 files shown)


jmfayard:
Looks promising to me. 

Can you add an example of use in the samples?
You can then try it with:

```bash
# Publish the latest version of the plugin to mavenLocal()
$ ./gradlew install  

# Try out the samples
$ ./gradlew -p examples/groovy dependencyUpdate
$ ./gradlew -p examples/kotlin dependencyUpdate
```

anuraaga:
Thanks - added examples

src/main/groovy/com/github/benmanes/gradle/versions/updates/Resolver.groovy
```diff
@@ -117,6 +118,15 @@ class Resolver {
       createQueryDependency(dependency, revision)
     }
 
+    // Common use case for dependency constraints is a java-platform BOM project.
+    try {
+      configuration.dependencyConstraints.each { dependency ->
+        latest.add(createQueryDependency(dependency, revision))
+      }
+    } catch (MissingPropertyException e) {
```
ben-manes: In other cases we've guarded using `metaClass.respondsTo("...")`. Which do you consider preferable?

src/main/groovy/com/github/benmanes/gradle/versions/updates/Resolver.groovy
```diff
@@ -149,6 +159,19 @@ class Resolver {
     }
   }
 
+  /** Returns a variant of the provided dependency used for querying the latest version. */
+  @TypeChecked(SKIP)
+  private Dependency createQueryDependency(DependencyConstraint dependency, String revision) {
```
ben-manes: Its probably not simpler to combine this with the other `createQueryDependency`, though the duplication could lead to mistakes of fixing one instead of both.

ben-manes:
Thanks for the PR and test case. I think it makes sense to go in. What do you think @jmfayard?

Would constraints someday be used to implement BOMs or are they totally separate? My only concern is if someone has a large BOM but only uses a few dependencies, and the constraint resolution causes them to see the universe. That seems unlikely but worth asking.

jmfayard:
I have no further remarks. I learned a lot of new things by reading the two links from the PR

anuraaga:
Switched to using `respondsTo`.

It's an interesting question of if in the future, Gradle would internally implement a BOM dependency by translating it into constraints. I feel as if this wouldn't happen, since a `Configuration` basically reflects the settings of the project and such opaque translation would probably not happen at that layer, BOM's work fine just using the normal dependency resolver already. I am more worried about if some other plugin, like `nebula`, does such translation as it could be a way to support BOM for old versions of Gradle that didn't natively resolve them - it's probably a bit too much though to worry about interactions with other plugins unless getting a report IMO.

ben-manes:
That makes sense. Thanks again. 😄

ben-manes:
Releasing v26.

PR #359, opened 2019-11-12
Work around Gradle 6.0 deprecations.

### LINKED ISSUE #357: Issues with Gradle 6.0-rc-1
When I use version _0.27.0_ with _Gradle 6.0 rc1_ I get the following output:

```
> Task :dependencyUpdates
Property 'checkConstraints' has redundant getters: 'getCheckConstraints()' and 'isCheckConstraints()'. This behaviour has been deprecated and is scheduled to be removed in Gradle 7.0.
Property 'checkForGradleUpdate' has redundant getters: 'getCheckForGradleUpdate()' and 'isCheckForGradleUpdate()'. This behaviour has been deprecated and is scheduled to be removed in Gradle 7.0.
Property 'outputFormatter' is not annotated with an input or output annotation. This behaviour has been deprecated and is scheduled to be removed in Gradle 7.0.
Property 'resolutionStrategy' is not annotated with an input or output annotation. This behaviour has been deprecated and is scheduled to be removed in Gradle 7.0.
```

anuraaga:
- Groovy generates both `getFoo` and `isFoo` for boolean properties and Gradle doesn't like it anymore. So manually define `getFoo` to disable Groovy's generation

- All properties must be annotated, as a reminder for task authors I guess. Annotated the properties that can be `Closure` as `@Internal`

- Unit test now ensure we don't use deprecated stuff

Fixes #357

### DIFF
--- a/src/main/groovy/com/github/benmanes/gradle/versions/updates/DependencyUpdatesTask.groovy
+++ b/src/main/groovy/com/github/benmanes/gradle/versions/updates/DependencyUpdatesTask.groovy
@@ -96,7 +96,7 @@ class DependencyUpdatesTask extends DefaultTask {
     }
 
     def evaluator = new DependencyUpdates(project, resolutionStrategyAction, revisionLevel(),
-      getOutputFormatter(), outputDirectory(), getReportfileName(), checkForGradleUpdate, gradleReleaseChannelLevel(),
+      outputFormatterProp(), outputDirectory(), getReportfileName(), checkForGradleUpdate, gradleReleaseChannelLevel(),
       checkConstraints)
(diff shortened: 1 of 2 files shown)


ben-manes:
Thank you!

I need to find the time to look into #334 before releasing. The advice of removing the workaround of disabling `canBeResolved` sounds promising. It just needs to be feature flagged smartly first.

juergenzimmermann:
@ben-manes Do you see a chance for releasing a new version?

ben-manes:
6.2 supposedly will break this plugin with a new feature, so waiting for that to release with a fix.

Issue #366, opened 2019-12-26
Review/support upcoming dependency verification in Gradle 6.2


vlsi:
Gradle 6.2 will have integrated dependency verification for integrity and provenance.

Issue: https://github.com/gradle/gradle/issues/10443
Documentation PR: https://github.com/gradle/gradle/pull/11755

Note: as Gradle would switch to "verify all resolutions", it would effectively prevent the operation of `gradle-versions-plugin`.

It looks like there will be `ResolutionStrategy#disableDependencyVerification()` API to temporary disable the verification.

Do you think that would be OK for G-V-P?

ben-manes:
Thanks for the heads up. If we use Groovy's `obj.metaClass.respondsTo` to reflectively check for that method, it could be added in a backwards compatible fashion.

anuraaga:
@vlsi I was looking at this now that 6.2 RC1 is released - did you mean this behavior would be automatically enabled or that a user may enable it manually and we need to account for it?

I added `6.2-rc-1` to the integration test and it passed without any code change so wondering if there's anything we actually need to do here.

https://github.com/ben-manes/gradle-versions-plugin/blob/master/src/test/groovy/com/github/benmanes/gradle/versions/DifferentGradleVersionsSpec.groovy#L118

anuraaga:
Looking more at the docs - I guess the test case needs to have the XML file included for any behavior change. Let me see if I can trigger it

https://github.com/gradle/gradle/pull/11755/files?short_path=b4b6735#diff-b4b6735dfc9950354518beb3ed1f4117

vlsi:
@anuraaga , it is not enabled by default.
It is activated by presence of `gradle/verification-metadata.xml` file.

For instance, if you launch Gradle with `--write-verification-metadata` it will create a file.
Then all subsequent Gradle executions would verify the dependencies and deny resolution of unresolved ones.

anuraaga:
@vlsi I merged the unit test in #375 and suspect we can close this issue - please close if you think we're all done here.

ben-manes:
I tried this with 6.2 and it fails when a `verification-metadata.xml` is present.

```
Dependency verification is an incubating feature.

FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring root project 'serverless'.
> Dependency verification failed for configuration ':classpath':
[...]

PR #375, opened 2020-02-07
Add test case that enables dependency verification.

anuraaga:
I believe I have enabled dependency verification by including the XML file and seeing a log message about it. Things still work fine. @vlsi can you check if this is the case you're concerned about?

For #366


src/test/groovy/com/github/benmanes/gradle/versions/DifferentGradleVersionsSpec.groovy
```diff
@@ -115,7 +115,8 @@ final class DifferentGradleVersionsSpec extends Specification {
       '5.4.1',
       '5.5.1',
       '5.6',
-      '6.0'
+      '6.0',
+      '6.1.1',
```
ben-manes: Shouldn’t you have 6.2-rc-1 here?

anuraaga: At first I did but realized I need a totally different test for the dependency verification. But yeah it doesn't hurt to check it here too.

ben-manes: Oh I see, I missed it being explicit in the new test

vlsi: Just in case: `dependency verification` is a silent feature in Gradle 6.1.
In other words, it is there, however, it is not documented.

In other words, there might be users who have `gradle/verification-metadata.xml` and Gradle 6.1 at the same time.

anuraaga:
I noticed the test maven repo doesn't have a guice 2.0 POM file in it so thought switching to 3.0 might bring up the potentially broken behavior, but even with 3.0, which does have a POM in the repo, still seems to work ok

ben-manes:
What about in an actual project? Can you get anything to break with this feature?

anuraaga:
Works fine I think - https://github.com/anuraaga/gradle-scratch

Actually, I could test out the verification itself since when I added the versions plugin, it complained about the versions plugin's checksum not being in the XML - expected behavior when checking checksums. Updated the XML and then proceeds fine.

```
Dependency verification is an incubating feature.

> Task :dependencyUpdates

 - com.github.ben-manes.versions:com.github.ben-manes.versions.gradle.plugin:0.27.0

 - com.google.inject:guice [3.0 -> 4.2.2]

 - Gradle: [6.2-rc-1: UP-TO-DATE]

Generated report file C:\tools\msys64\home\Anuraag\git\gradle-scratch\build\dependencyUpdates\report.txt

Deprecated Gradle features were used in this build, making it incompatible with Gradle 7.0.
Use '--warning-mode all' to show the individual deprecation warnings.
See https://docs.gradle.org/6.2-rc-1/userguide/command_line_interface.html#sec:command_line_warnings

BUILD SUCCESSFUL in 5s
1 actionable task: 1 executed
```

@vlsi Does it make sense that the versions plugin works fine even with verification enabled? If so, glad we won't have any issues with 6.2 :)

ben-manes:
Maybe it’s because we only resolve metadata and not any artifacts, so the checksum isn’t triggered?

vlsi:
@melix, can you please double-check the dependency verification case here?

It looks like the verification does not stop resolution of detached configurations which seems odd to me.

Just in case, I've checked with Caffeine sources as well, and `dependencyUpdates` task works even after `--write-verification-metadata sha256` (Gradle 6.1.1)

melix:
It's hard to figure out what you mean without an example. Dependency verification should be tested with 6.2 only (not 6.1). Verification of detached configurations should always happen, unless it's disabled in the verification configuration file.

anuraaga:
@melix @vlsi I realized that this probably works since we resolve new versions using a dynamic version number

https://github.com/ben-manes/gradle-versions-plugin/blob/master/src/main/groovy/com/github/benmanes/gradle/versions/updates/Resolver.groovy#L151

I presume verification isn't applied to dynamic versions since it's not clear what the checksum should actually refer to.

Does this sound right? And if so I guess it's why we don't have to worry about the effect of verification on this plugin?

melix:
Verification happens once the artifact or metadata is accessed. So it works for dynamic versions _too_ (once resolved to a version number). It doesn't, however, apply to _changing_ versions (like snapshots) because by definition their checksum would always change.

anuraaga:
I see. Well, the line of code that resolves a bunch of dependies with something like version `latest` is here

https://github.com/ben-manes/gradle-versions-plugin/blob/master/src/main/groovy/com/github/benmanes/gradle/versions/updates/Resolver.groovy#L84

These newer versions that get resolved to don't appear in our XML and I guess we are expecting a verification failure. That being said, we don't use the resolved artifacts at all, we only do this to get resolved version numbers from Gradle, not to actually use any dependency. So perhaps we bail out of a flow before verification triggers.

I guess knowing which exact methods have a chance of producing a verification failure would help clear our confusion.

anuraaga:
Thanks - guess we can be confident that Gradle 6.2 will still work with versions plugin :)

Issue #127, opened 2017-02-01
Support dependency configurations from java-library


pkubowicz:
Gradle 3.4 RC introduces 'java-library' as a replacement for 'java'. Running 'dependencyUpdates' on such a project produces errors:
```
Failed to resolve ::api
Failed to resolve ::apiElements
Failed to resolve ::implementation
Failed to resolve ::runtimeElements
Failed to resolve ::runtimeOnly
Failed to resolve ::testImplementation
Failed to resolve ::testRuntimeOnly
```

Also, the plugin fails to check the new configurations.

To reproduce use `build.gradle` with
```
plugins {
    id 'com.github.ben-manes.versions' version '0.13.0'
    id 'com.github.johnrengelman.shadow' version '1.2.3'
}

apply plugin: 'java-library'

repositories {
    jcenter()
}

dependencies {
    api 'org.slf4j:slf4j-api:1.7.0'
    implementation 'com.google.guava:guava:19.0'
    compileOnly 'javax.servlet:javax.servlet-api:3.0.1'
    runtimeOnly 'ch.qos.logback:logback-classic:1.1.0'

    testImplementation 'junit:junit:4.10'
    testCompileOnly 'org.projectlombok:lombok:1.16.0'
    testRuntimeOnly 'org.slf4j:log4j-over-slf4j:1.7.0'
}
```

Each dependency in this file (except for the versions plugin) is outdated and should be reported. Currently only plugins, `compileOnly` and `runtimeOnly` are checked.

ben-manes:
This is because Gradle disallows resolving those configurations (see #125). I don't know why they did that. I'm hoping that cloning and setting the flag to allow will bypass this. We should probably ask to find out why, in case they'd prefer to revert in rc2/

pkubowicz:
Have you asked them? Do you need some help?

ben-manes:
[Yep](https://discuss.gradle.org/t/gradle-3-4-rc1-is-now-available-for-testing/21351/7?u=benjamin_manes). I'm a little concerned that `validateMutation` will reject it, but I haven't tested it yet.

ben-manes:
Adding @jochenberger, @oehme

Two approaches seem to work. Both seem a little hacky.

__Disable Restriction__
```
configuration = configuration.copyRecursive().setTransitive(false)
configuration.setCanBeResolved(true)
```
[...]

PR #378, opened 2020-02-21
Upgrade deps and use @CompileStatic everywhere

### LINKED ISSUE #365: Gradle 6 deprecation warnings
When running `dependencyUpdates` on Gradle 6, the following warnings appear:

```
Property 'checkConstraints' has redundant getters: 'getCheckConstraints()' and 'isCheckConstraints()'. This behaviour has been deprecated and is scheduled to be removed in Gradle 7.0.
Property 'checkForGradleUpdate' has redundant getters: 'getCheckForGradleUpdate()' and 'isCheckForGradleUpdate()'. This behaviour has been deprecated and is scheduled to be removed in Gradle 7.0.
Property 'outputFormatter' is not annotated with an input or output annotation. This behaviour has been deprecated and is scheduled to be removed in Gradle 7.0.
Property 'resolutionStrategy' is not annotated with an input or output annotation. This behaviour has been deprecated and is scheduled to be removed in Gradle 7.0.
```

### LINKED ISSUE #377: Support Gradle 6.2 (due to changes in org.gradle.util.SingleMessageLogger)
When running on Gradle 6.2, the plugin fails with:

```
Caused by: groovy.lang.MissingMethodException: No signature of method: static org.gradle.util.SingleMessageLogger.nagUserWith() is applicable for argument types: (String, String, String, null) values: [dependencyUpdates.resolutionSt
rategy, , Remove the assignment operator, '=', when setting this task property, ...]
        at com.github.benmanes.gradle.versions.updates.DependencyUpdatesTask.dependencyUpdates(DependencyUpdatesTask.groovy:81)
[...]
```

The Gradle class [SingleMessageLogger](https://github.com/gradle/gradle/blob/master/subprojects/logging/src/main/java/org/gradle/util/SingleMessageLogger.java) has been rewritten, and should probably not be used anymore (it looks like it is only there for backwards compatibility with the Nebula plugin).

The only replacement is `org.gradle.internal.deprecation.DeprecationLogger`, which, being an internal class, is not great either.

SUPERCILEX:
This PR:
- Requires users have at least Gradle 5.0 (I kept getting Groovy linking issues on older versions. Not sure what your policy on minimum dependencies is, but Gradle 5.0 was released way back in 2018, so I feel like killing 3.x and 4.x is warranted.)
- Fixes #377, closes #365


SUPERCILEX:
Plugin application perf improved by about 10X: [old](https://scans.gradle.com/s/7zdwtysr2fcry/performance/configuration#com.github.ben-manes.versions) -> [new](https://scans.gradle.com/s/zh4xis5pq2azk/performance/configuration?openScriptsAndPlugins=WzIsMSwwXQ#com.github.ben-manes.versions).

src/main/groovy/com/github/benmanes/gradle/versions/updates/Resolver.groovy
```diff
@@ -65,7 +67,7 @@ class Resolver {
     this.projectUrls = new ConcurrentHashMap<>()
     this.resolutionStrategy = resolutionStrategy
     this.project = project
-    this.checkConstraints = checkConstraints;
+    this.checkConstraints = checkConstraints
 
     useSelectionRules = new VersionComparator(project)
```
ben-manes: I suppose these workarounds could go away if 5.0+ as a baseline (we started this plugin in 1.0 world!)

SUPERCILEX: Damn, that's ancient! 🤯 Does caa6aa2ac135aa848f8523f3dcd535c31af35c0f look good? And is there anything else we can remove?

ben-manes:
Possibly we can condition the `canBeResolved` hack? This was added when ensuring compatibility with `compile` and 6.0 strongly encourages a transition to `implementation` and friends. In #334 there is a belief that us forcing it to be enabled breaks the newish Kotlin platform. If that would fix it, we could be gentle by not supporting `compile` in 6.2+ and assume those on earlier versions will migrate. In 7.0 they are being removed, so it was a 5.0 compatibility issue when first introduced.

SUPERCILEX:
No, that doesn't seem to fix the issue (I reproed using Zac's app). It also broke the plugin because the resolved dependency set was empty.

ben-manes:
Did you try skipping the configurations where `canBeResolved` is true, instead of us trying to force resolution? That's the fix I had in mind, as those user-defined configurations are deprecated. That could be done prior to any resolution logic, e.g. at https://github.com/ben-manes/gradle-versions-plugin/blob/4a4ee270645b3ba23155c121634c5b93ebec226a/src/main/groovy/com/github/benmanes/gradle/versions/updates/DependencyUpdates.groovy#L66-L74

I was hoping that just ignoring those configuration types, like `compile`, would be okay in 6.2+ now and also fix their Kotlin problems. I just haven't had the time to help work on this plugin.

SUPERCILEX:
Huh, I clearly don't understand how configuration works at all. I would have thought we'd want to skip `canBeResolved = false`. Gonna read up on this stuff and report back. 👍

ben-manes:
Sorry, I meant skip `canBeResolved = false` like you said. :)

SUPERCILEX:
Well yes, but no. 😁 Nothing I've tried fixes the issue where MPP deps can't be resolved. That said, only processing `config.canBeResolved == false || config.canBeConsumed == true` does actually fix that annoying issue where a ton of `Failed to resolve ...` logs are spewed. I had to special case annotation processors, but I feel like that quirk is worth it, especially since it also improves performance (~35% speedup on CatchUp: [old](https://scans.gradle.com/s/fjamyeycguofk/timeline?details=6liaj6impyc5y) -> [new](https://scans.gradle.com/s/mc6lzlx7xiikk/timeline?details=6liaj6impyc5y)).

SUPERCILEX:
From what I understand, this is what we're doing:
- We process `canBeResolved == false` configurations.
- That means things like `implementation` but not `compileClasspath`. This is actually good because it means we're only processing configurations users have explicitly set up.
- When we resolve the configuration, we mark it as resolvable. How or why this works, I don't know.

ben-manes:
We marked it as resolvable because most people used `compile` prior to Gradle 6. When it became unresolvable, then the plugin would have forced everyone to upgrade to the newest convention. Only in 7.x is the classic way finally gone, which means Gradle gave us major 4 releases to get there. In 6.x those are deprecated, we could assume everyone migrates, and don't need to hack the configuration anymore.

Would you prefer retaining the old hack for 5.x users and only applying your `isUsefulConfiguration` at a cutoff version,  6.2+? Or would you like to drop the hacks and assume everyone should have updated their builds to adopt the latest version of this plugin. I think either is a reasonable choice.

SUPERCILEX:
It doesn't seem to be a hack though, we actually _want_ configurations with `canBeResolved == false` and then we have to mark them as resolvable (or at least least when I comment out those two spots added in #127, the plugin doesn't find any of the dependencies).

ben-manes:
Oh okay. I suppose its been so many workarounds and Gradle versions I no longer can tell how anything will work. 😄 

When you're happy, feel free to merge. You are also welcome to cut a release:
1. Increment version in build.gradle and commit
2. Add Github release notes
3. CI will see the new tag and deploy the archive.

SUPERCILEX:
Dope, will do!

SUPERCILEX:
@ben-manes Do I have to do anything to make it show up on https://plugins.gradle.org?

anuraaga:
@SUPERCILEX Thanks for the help with this!

Just a strawman's proposal, would it be reasonable to go ahead and migrate the code to Java? I enjoy helping to maintain this project, but do sometimes feel a little down when dealing with a Groovy gotcha. Not enough to not do it though, but I believe @ben-manes is also a Java person so maybe it's a good fit? If so, I'm happy to do the gruntwork

ben-manes:
@SUPERCILEX Nope. It will sync, but is slow. We were early adopters of the plugin portal which at that time went through this sync phase, whereas perhaps now you can deploy to directly? All I know is it eventually works 😄 

@anuraaga Groovy is certainly not my language and I used what Gradle expected at that time. I would be supportive of whatever current maintainers prefer (Java, Kotlin, etc). I'm no longer active except as a facilitator and will happily let those willing to do the work make the decisions.

SUPERCILEX:
@anuraaga Yeah, I'd actually love to rewrite this plugin in Kotlin at some point—I just don't have the time 😅. I'd be willing to take a look at pull requests though (either Kotlin or Java) if somebody else has the time.

@ben-manes Lol, SGTM. I didn't know you could sync stuff to the plugin portal. And yeah, I deploy my plugins to the portal directly.

bishiboosh:
Don't know if I will have the time to do it in one setting but I'm willing to try to rewrite the plugin in Kotlin

PR #385, opened 2020-03-05
Upgrade documentation with --refresh-dependencies flag

benckx:
I kept it simple. Personally I searched for "refresh" on the README. I suppose other people might have the same issue.


ben-manes:
Thanks!

PR #392, opened 2020-04-16
Add hint about how to add the plugin via an init script

jochenberger:
With a certain amount of Gradle projects, I became too lazy to add the plugin to each and every one of them. This is the solution I came up with, so I thought I'd share it. :-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,28 @@ buildscript {
 }
 ```
 
+### using a Gradle init script ###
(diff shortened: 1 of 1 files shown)


ben-manes:
you don't need me to review your PRs! =)

jochenberger:
When it comes to documentation, it's sometimes better to have a native speaker have a closer look. 😉

PR #456, opened 2021-01-07
Move to Github Actions

jaredsburrows:
Remove TravisCi in favor of Github Actions

- Added `build.yml` for building and testing on git pushes and Github pull requests
- Added `deploy.yml` for publishing the plugin when new tags are created


.github/workflows/build.yml
```diff
@@ -0,0 +1,57 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        java-version:
+          - 1.8
+          - 11
+
+    steps:
+      - name: Cache Gradle dependencies
+        uses: actions/cache@v2
```
anuraaga, line 21: I'd generally recommend Vlad's excellent native-cache plugin

https://github.com/burrunan/gradle-cache-action

example: https://github.com/open-telemetry/opentelemetry-java/blob/master/.github/workflows/pr-build.yml#L36

.github/workflows/build.yml
```diff
@@ -0,0 +1,57 @@
+name: build
+
+on: [push, pull_request]
+
+env:
+  GRADLE_OPTS: "-Dorg.gradle.jvmargs=-Xmx4g -Dorg.gradle.daemon=false -Dkotlin.incremental=false"
```
anuraaga, line 6: I think Gradle's recommendation is to use daemon in CI too nowadays so maybe don't need this

.github/workflows/build.yml
```diff
@@ -0,0 +1,57 @@
...
+          name: gradle-versions-plugin-${{ github.workflow }}-${{ github.run_id }}
+          path: |
+            build/libs
+            build/outputs
+            build/publications
+            build/distributions
+            build/reports
+            build/test-results
+          if-no-files-found: warn
+
```
anuraaga, line 57: Looks like `bintrayUpload` for master build is missing? Presumably requires credentials to set up but until it, I don't think `.travis.yml` can be removed.

ben-manes:
Oh thanks! I meant to do this but didn't get to it. @jochenberger was nice and dug up the bintray api key, which we set as a github secret env variable `BINTRAY_API_KEY`.

I don't think this mirrors the travis ci script, though. That will perform a deployment to bintray when a new tag is created. I believe we need that step to automate the deployment

jaredsburrows:
@ben-manes We can add it as a github secret. Do you want this publishing on each merge?

.github/workflows/build.yml
jaredsburrows: Did we want this in the master build?

.github/workflows/build.yml
jaredsburrows: Want me to just copy/paste this?

anuraaga: Well I think the properties / arguments will differ a bit but yeah should mostly be copy-paste :)

jaredsburrows: Why use this? Can we specify paths?

anuraaga: The README has a description of why to use it :) Don't need to specify paths since it's Gradle-aware and does it properly automatically

.github/workflows/build.yml
jaredsburrows: Done.

.github/workflows/build.yml
jaredsburrows: Done.

jaredsburrows:
@ben-manes Can you add the `BINTRAY_API_KEY` and `BINTRAY_USERNAME` to the "secrets" of this repository?

.github/workflows/build.yml
jaredsburrows: So you want me to replace the build "step" and current cache plugin with this?

anuraaga: Yeah

ben-manes:
@jaredsburrows Added. We want to publish a new release when explicitly requested. Right now that is done when a new tag is created, which worked pretty well. Do you have a preference for an alternative approach, e.g. by checking for a tag string on a commit message?

ben-manes:
Note that you'll need to update the `build.gradle` section to use these new env variables.

anuraaga:
@ben-manes It's not necessarily better but as an example, another approach could be to have an input box with the version and let the workflow create the tag as well

https://github.com/open-telemetry/opentelemetry-java/blob/master/.github/workflows/release-build.yml#L8

ben-manes:
Oh that is neat. I didn't know about inputs on actions. The tag approach is nice as it happens automatically when writing the github release notes. I'm sure with the inputs you could get fancy by automatically updating the build's version string. That's all overkill here, but it's nifty to see how actions are such a game changer.

jaredsburrows:
@ben-manes I have updated the environment variables in `build.gradle`.

.github/workflows/build.yml
jaredsburrows: Done.

jaredsburrows: Seems to have a problem on the CI?
```
Warning: Gradle wrapper configuration is not found at /home/runner/work/gradle-versions-plugin/gradle-versions-plugin/gradle/wrapper/gradle-wrapper.properties.
Will use the current release Gradle version
/usr/bin/git log -1 --quiet --format=%T
fatal: not a git repository (or any of the parent directories): .git
(node:2407) UnhandledPromiseRejectionWarning: Error: The process '/usr/bin/git' failed with exit code 128
    at ExecState._setResult (/home/runner/work/_actions/burrunan/gradle-cache-action/v1.6/dist/cache-action-entrypoint.js:45001:468)
    at ExecState.CheckComplete (/home/runner/work/_actions/burrunan/gradle-cache-action/v1.6/dist/cache-action-entrypoint.js:44994:45)
    at ChildProcess.<anonymous> (/home/runner/work/_actions/burrunan/gradle-cache-action/v1.6/dist/cache-action-entrypoint.js:44961:84)
    at ChildProcess.emit (events.js:210:5)
    at maybeClose (internal/child_process.js:1021:16)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:283:5)
```

ben-manes:
lgtm assuming all the kinks are worked out. Deferring to @anuraaga to approve. Thanks guys! 🙂

jaredsburrows:
@ben-manes @anuraaga To make this cleaner and shorter, we can do something like this: https://github.com/JakeWharton/gitout/blob/master/.github/workflows/build_and_publish.yml. What do you all think?

ben-manes:
sure, it your guys call

.github/workflows/build.yml
```diff
@@ -0,0 +1,49 @@
...
+      matrix:
+        java-version:
+          - 1.8
+
+    steps:
+      - name: Cache Gradle dependencies
+        uses: burrunan/gradle-cache-action@v1.6
+        with:
+          remote-build-cache-proxy-enabled: false
+          arguments: clean build -s --scan
```
anuraaga, line 20:
```suggestion
```

This actually runs the build with caching enabled, we don't need to run twice

.github/workflows/build.yml
```diff
@@ -0,0 +1,49 @@
...
+      - name: Validate Gradle Wrapper
+        uses: gradle/wrapper-validation-action@v1
+
+      - name: Configure JDK
+        uses: actions/setup-java@v1
+        with:
+          java-version: ${{ matrix.java-version }}
+
+      - name: Run Build
+        run: ./gradlew clean build -s --scan
```
anuraaga, line 34:
```suggestion
      - name: Run build with caching enabled
        uses: burrunan/gradle-cache-action@v1.6
        with:
          remote-build-cache-proxy-enabled: false
          arguments: clean build -s --scan
```

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,30 @@
...
+      - name: Configure JDK
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+
+      - name: Publish
+        run: ./gradlew clean build bintrayUpload -s --scan
+        env:
+          BINTRAY_API_KEY: ${{ secrets.BINTRAY_API_KEY }}
+          BINTRAY_USERNAME: ${{ secrets.BINTRAY_USERNAME }}
```
anuraaga, line 29:
```suggestion
      - name: Publish
        uses: burrunan/gradle-cache-action@v1.6
        with:
          remote-build-cache-proxy-enabled: false
          arguments: clean build bintrayUpload -s --scan
        env:
          BINTRAY_API_KEY: ${{ secrets.BINTRAY_API_KEY }}
          BINTRAY_USERNAME: ${{ secrets.BINTRAY_USERNAME }}
```

Otherwise it'll have to download all the dependencies every time

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,30 @@
...
+
+on:  
+  push:
+    tags:
+      - '*'
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/master'
```
anuraaga, line 11: I'm presuming we'd only tag if we need to release - so in the off chance the tag was a release branch or something like that, may as well publish it.

anuraaga (commented):
I personally don't mind multiple files to reduce the cognitive load of if statements in yaml files, but don't have a strong preference on it.

.github/workflows/deploy.yml
jaredsburrows: So remove this if check?

anuraaga: I think so, especially if @ben-manes agrees :)

anuraaga (approved):
LGTM pending last point to confirm about the tags branch

.github/workflows/deploy.yml
ben-manes: Makes sense

jaredsburrows: Done.

jaredsburrows:
@ben-manes @anuraaga Updated the PR. Letting the build run now. Let me know if you want anything else done in this PR.

ben-manes:
I think you can merge once it passes. Thanks for doing this!

jaredsburrows:
@ben-manes @anuraaga Build passed. Merging now!

PR #459, opened 2021-01-09
Use Gradle Repo instead of Jcenter

jaredsburrows:
- Use Gradle Repo - mirrors `jcenter` and `mavenCentral`

Helps to address @anuraaga comment here: https://github.com/ben-manes/gradle-versions-plugin/pull/457#discussion_r554291918.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,6 @@
 buildscript {
   repositories {
-    jcenter()
+    gradlePluginPortal()
   }
 
   dependencies {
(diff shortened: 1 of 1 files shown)


ben-manes:
Gradle repo is hosted by jcenter, fyi 

only the plugins should use it, not regular dependencies

jaredsburrows:
@ben-manes In that case, I will close this PR.

PR #460, opened 2021-01-09
Update publishing and remove nexus plugin

jaredsburrows:
- Remove `gradle-nexus-plugin` plugin
 - Apply the `com.gradle.plugin-publish` plugin
 - Add `sourcesJar`, `testsJar` and `reportsZip` in addition to `docsJar`
 - Update POM and add `issueManagement`
 - Add more metadata for `bintray`(Jcenter Bintray) and `pluginBundle`(Gradle Plugins)

To verify POM and artifact creation you can run `gradlew publish` and if you want to test `bintray`, you can run `gradlew bintrayUpload` but you need to add `dryrun = true` to `brintray`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,25 +1,14 @@
-buildscript {
-  repositories {
-    jcenter()
-  }
-
-  dependencies {
-    classpath 'org.gradle.api.plugins:gradle-nexus-plugin:0.7.1'
-  }
-}
-
 plugins {
-  id "com.jfrog.bintray" version "1.8.4"
-  id "java-gradle-plugin"
-  id "groovy"
-  id "codenarc"
+  id 'com.jfrog.bintray' version '1.8.4'
+  id 'java-gradle-plugin'
+  id 'maven-publish'
+  id 'groovy'
+  id 'codenarc'
 }
 
 repositories {
   jcenter()
 }
-
-apply plugin: 'nexus'
 
 group = 'com.github.ben-manes'
 version = '0.36.0'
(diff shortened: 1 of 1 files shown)


ben-manes:
Do we still use that? A long time ago this plugin was on Central, but then switched to JCenter as that was how the plugin portal worked in its early days. Other than defining the pom as a dsl (which the newer maven plugin mirrors), I don't think it is used.

jaredsburrows:
You are right. We are using `com.jfrog.bintray`. Can we remove this plugin and just use a custom POM like I do here: https://github.com/jaredsburrows/gradle-license-plugin/blob/master/gradle/publish.gradle.

ben-manes:
Yes, let's do that. Since it uses the legacy maven task that plugin isn't compatible with the latest Gradle versions, too.

Another reference, if helpful, is
https://github.com/ben-manes/caffeine/blob/master/gradle/publish.gradle

jaredsburrows:
Thanks for that example. Lets go with that. Did you also want to support snapshots and maven central?

ben-manes:
Your's is nice too, don't feel the need to follow mine.

I don't think the snapshots is necessary

jaredsburrows:
@ben-manes Bintray publishing plugin actually has a maven sync, https://github.com/bintray/gradle-bintray-plugin#buildgradle. Have you used this before?

ben-manes:
I knew of it but have not used it. I think the plugin portal is enough. The bintray repository is owned by a former contributor and not worth bothering him for a sync to Central.

jaredsburrows:
@ben-manes Updated. When are you uploading the gradle plugins? It was not in the build.gradle.

ben-manes:
When it is uploaded to bintray then it syncs automatically. This is how the plugin portal originally worked and only much later was a direct upload capability supported.

ben-manes (approved):

jaredsburrows:
@ben-manes I added more metadata for bintray(Jcenter Bintray) and pluginBundle(Gradle Plugins) for https://bintray.com/beta/#/fooberger/maven/com.github.ben-manes:gradle-versions-plugin?tab=overview and https://plugins.gradle.org/plugin/com.github.ben-manes.versions.

jaredsburrows:
@ben-manes I have reviewed this and tested the local generated POM and outputs. Should I remove `pluginBundle` since you do not manually upload to Gradle plugins repo. I am thinking that adding extra metadata to `bintray`, would the Gradle plugin repo sync the meta data? If so, I can remove `pluginBundle` and `com.gradle.plugin-publish`.

ben-manes:
You can probably remove it if you prefer and most likely the metadata passes through. Thank you for modernize the build. 🙂

PR #467, opened 2021-01-14
Bump gradle to 6.8

Goooler:
Gradle 6.8 enhanced Kotlin DSL script performance, see [Release Note](https://docs.gradle.org/6.8/release-notes.html) here

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.3-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.8-all.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists


ben-manes:
Looks like it failed the tests. [Gradle 6.5](https://docs.gradle.org/6.5/release-notes.html#dependency-management-improvements) changed its version detection logic (for the worse in our use-case) and that might be the culprit. You'd have to either debug the test cases or, if not enough interest, close this issue.

Goooler:
> Looks like it failed the tests. [Gradle 6.5](https://docs.gradle.org/6.5/release-notes.html#dependency-management-improvements) changed its version detection logic (for the worse in our use-case) and that might be the culprit. You'd have to either debug the test cases or, if not enough interest, close this issue.

Ok

ben-manes:
Oh, it looks like @jaredsburrows hard coded the gradle upgrade string.

```
  - Gradle: [6.3 -> 6.8]
```

I think that is an oversight in those new tests, because it will fail whenever a new release occurs. This failed now because the base version is upgraded.

I think the right fix here is to remove that using `checkForGradleUpdate = false` so that the test doesn't check the gradle version.

anuraaga:
I think when updating Gradle we need to add more test versions here too

https://www.github.com/ben-manes/gradle-versions-plugin/tree/master/src%2Ftest%2Fgroovy%2Fcom%2Fgithub%2Fbenmanes%2Fgradle%2Fversions%2FDifferentGradleVersionsSpec.groovy

Or the new Gradle might break sometimes on one or the unrested Gradle versions

jaredsburrows:
@ben-manes @anuraaga PR to fix this - https://github.com/ben-manes/gradle-versions-plugin/pull/468

jaredsburrows:
Can this PR be updated (with the fix)? This PR needs to run `gradlew wrapper` to update the bat/bash scripts.

jaredsburrows:
@Goooler Thanks for sending this in! I really appreciate it but I needed to fix/future poof a test for this PR to work. I did this in #469. Looking forward to your future contributions!

Goooler:
> @Goooler Thanks for sending this in! I really appreciate it but I needed to fix/future poof a test for this PR to work. I did this in #469. Looking forward to your future contributions!

Thx

Issue #477, opened 2021-02-17
Show "reason" in dependency report


alkamo:
### Feature Request

Since 4.6, Gradle has supported the ability to populate a "reason" attribute for a dependency. This would be useful to display in the dependency update report, since it may offer a justification for using an older version and hint at when it might become updateable.

### Proposed functionality

In response to the following configuration:
```
dependencies {
    implementation 'net.java.dev.jna:jna:3.2.7' {
        reason "Versions above 3.2.7 will not work on RHEL6"
    }
}
```

The dependency update task would display:
```
The following dependencies have later release versions:
 - net.java.dev.jna:jna [3.2.7 -> 5.7.0]
     Versions above 3.2.7 will not work on RHEL6
     https://github.com/java-native-access/jna
```

### Implementation Notes

The "reason" is set using the "becauseOf" method and fetched using the "reason" method.

alkamo:
I've looked into implementing this, but it's not nearly as straightforward as I had hoped.

- `Coordinates` is the only class that gets data populated from Gradle's `Dependency` object. Adding `reason` to that is easy enough (excluding it from `@EqualsAndHashCode`). However, this strikes me as a little weird, since it involves adding non-key data to a class that's intended to be a key.
- Since the set of `Coordinates` is translated into a `Map<Map<String, String>, String>` when it's passed to `DependencyUpdatesReporter`, there doesn't seem to be any good way to pass any additional data to `DependencyUpdatesReporter` under the current design.
- The most obvious thing to do seems to be to change the data passed to `DependencyUpdatesReporter` to `Map<Map<String, String>, Coordinate>`, but that's a pretty big change, which would have other downstream affects (plus, it seems to subvert the intention of the design of `DependencyUpdatesReporter`).

ben-manes:
Yeah, it got messy and the strings were an adapter between the two chunks of code which evolved a bit independently. It has needed a refactoring to clean that up for a while now. I’m okay if you want to take a stab at it, there was just no good reason before.

In your case, I think that I’ve used `force` to ensure no upgrades which is a bit heavy handed.

ben-manes:
Thanks!

PR #479, opened 2021-02-23
User defined reason

alkamo:
https://github.com/ben-manes/gradle-versions-plugin/issues/477

My primary objective was to add the "reason" that can be supplied by the user in a dependency declaration in the versions report. This turned out to be a more significant change than I expected and I made a bunch of associated changes along the way.

The biggest change was to make DependencyUpdatesReporter work with maps of Coordinates, instead of Strings. This allows the "reason" to get passed alongside the version.

I also changed all of the multi-line strings in PlainTextReporter to use one line per println statement. This was in response to trying to write a test for the plain text output including "reasons", only to discover that the plain-text output contained both CRLF and LF (without CR). By using println instead of embedded linefeeds, the system-appropriate line separator should be used consistently.


src/main/groovy/com/github/benmanes/gradle/versions/reporter/PlainTextReporter.groovy
```diff
@@ -58,18 +57,19 @@ class PlainTextReporter extends AbstractReporter {
   }
 
   private def writeHeader(printStream) {
-    printStream.println """
-      |------------------------------------------------------------
```
alkamo, line 62: As mentioned in the description, the embedded linefeeds caused inconsistent behavior on Windows. Splitting these into multiple println statements, should make the output consistent independent of OS.

src/main/groovy/com/github/benmanes/gradle/versions/reporter/result/Dependency.groovy
```diff
@@ -15,4 +15,5 @@ class Dependency {
   String name
   String version
   String projectUrl
+  String userReason
```
alkamo, line 18: I used "UserReason" because "Reason" was already being used for unresolved dependencies.

src/main/groovy/com/github/benmanes/gradle/versions/updates/Coordinate.groovy
```diff
@@ -54,24 +57,41 @@ class Coordinate implements Comparable<Coordinate> {
     return (result == 0) ? version.compareTo(coordinate.version) : result
   }
 
+  static Coordinate from(ExternalModuleDependency dependency) {
```
alkamo, line 60: This method was needed because ExternalModuleDependency implements both `ModuleVersionSelector` and `Dependency`, so it was going to the wrong method unless a more specific one was available.

src/main/groovy/com/github/benmanes/gradle/versions/updates/DependencyUpdatesReporter.groovy
```diff
@@ -175,27 +176,29 @@ class DependencyUpdatesReporter {
     )
   }
 
+  private static void updateKey(Map<String, String> existingKey) {
+    int index = existingKey['name'].lastIndexOf('[')
+    existingKey['name'] = (index == -1) ? existingKey['name'] : existingKey['name'].substring(0, index)
+  }
+
```
alkamo, line 183: It's only two lines, but the fact that this was repeated three times bothered me, so I abstracted it.

ben-manes (approved):
This looks great to me, thanks!

src/main/groovy/com/github/benmanes/gradle/versions/updates/DependencyUpdatesReporter.groovy
```diff
@@ -226,24 +228,28 @@ class DependencyUpdatesReporter {
     new DependenciesGroup<T>(dependencies.size(), dependencies)
   }
 
-  protected def buildDependency(String group, String name, String version, String projectUrl) {
-    new Dependency(group, name, version, projectUrl)
+  protected def buildDependency(Coordinate coordinate, Map<String, String> key) {
+    new Dependency(key['group'], key['name'], coordinate.getVersion(), projectUrls[key],
+      coordinate.getUserReason())
   }
```
alkamo, line 234: These builders were just acting as passthroughs for the constructor. Given that, it made sense to me to pass fewer arguments in and let the builders translate those arguments into the constructor parameters.

ben-manes:
Can you look into the Java 11 failures? I was able to reproduce locally.

src/test/groovy/com/github/benmanes/gradle/versions/DependencyUpdatesSpec.groovy
```diff
@@ -120,24 +121,10 @@ final class DependencyUpdatesSpec extends Specification {
...
-      ['group': 'backport-util-concurrent', 'name': 'backport-util-concurrent-java12']: '3.1',
-    ]
-    reporter.downgradeVersions == [
-      ['group': 'com.google.guava', 'name': 'guava']      : '99.0-SNAPSHOT',
-      ['group': 'com.google.guava', 'name': 'guava-tests']: '99.0-SNAPSHOT',
-    ]
+    checkUnresolvedVersions(reporter)
+    checkUpgradeVersions(reporter)
+    checkUpToDateVersions(reporter)
+    checkDowngradeVersions(reporter)
```
alkamo, line 127: Because these collections are now objects, doing a direct comparisons of the collections was no longer practical. I revised these tests to instead check the size of the collection and individual entries using the appropriate keys.

I also abstracted them, rather that update one and copy/paste it five times.

src/test/groovy/com/github/benmanes/gradle/versions/DependencyUpdatesSpec.groovy
```diff
@@ -636,28 +579,84 @@ final class DependencyUpdatesSpec extends Specification {
...
-      unresolvable 'com.github.ben-manes:unresolvable:1.0',
-        'com.github.ben-manes:unresolvable2:1.0'
+      upToDate('backport-util-concurrent:backport-util-concurrent:3.1') { because 'I said so' }
+      upToDate('backport-util-concurrent:backport-util-concurrent-java12:3.1')
+      exceedLatest('com.google.guava:guava:99.0-SNAPSHOT') { because 'I know the future' }
+      exceedLatest('com.google.guava:guava-tests:99.0-SNAPSHOT')
+      upgradesFound('com.google.inject:guice:2.0') { because 'That\'s just the way it is' }
+      upgradesFound('com.google.inject.extensions:guice-multibindings:2.0')
+      unresolvable('com.github.ben-manes:unresolvable:1.0') { because 'Life is hard' }
+      unresolvable('com.github.ben-manes:unresolvable2:1.0')
```
alkamo, line 603: In order to provide "reason" entries, I had to declare the dependencies slightly differently. Rather than only update the ones I needed to, I changed them all for consistency.

src/test/groovy/com/github/benmanes/gradle/versions/DependencyUpdatesSpec.groovy
```diff
@@ -636,28 +579,84 @@ final class DependencyUpdatesSpec extends Specification {
...
     }
   }
+
+  private static void checkUnresolvedVersions(def reporter) {
+    Map<Map<String, String>, ModuleVersionSelector> unresolvedMap = reporter.unresolved
+      .collect { it.selector }.collectEntries { dependency ->
+      [['group': dependency.group, 'name': dependency.name]: dependency]
+    }
+    assert reporter.unresolved.size() == 2
```
alkamo, line 612: When resolving a test comparison in a separate method, Spock needs the assert keyword.

src/test/groovy/com/github/benmanes/gradle/versions/OutputFormatterSpec.groovy
```diff
@@ -165,7 +165,7 @@ final class OutputFormatterSpec extends Specification {
     The following dependencies have later milestone versions:
      - com.google.inject:guice [2.0 -> 3.1]
          http://code.google.com/p/google-guice/
-      """.stripIndent()
+      """.stripIndent().replace('\r','').replace('\n', System.lineSeparator())
```
alkamo, line 168: This combination of `replace` methods should make this test OS-independent.

alkamo:
It's fixed.

ben-manes:
thanks! I'll try to release this, but it might be quirky since infra changed under us. We haven't released since migrating off TravisCI's shutdown, and haven't off JCenter which is about to shutdown. So yeah, 🤞

src/test/groovy/com/github/benmanes/gradle/versions/DependencyUpdatesSpec.groovy
```diff
@@ -627,7 +570,7 @@ final class DependencyUpdatesSpec extends Specification {
     new DependencyUpdates(project, resolutionStrategy, revision, outputFormatter, outputDir, reportfileName, checkForGradleUpdate, gradleReleaseChannel).run()
   }
 
-  private def addRepositoryTo(project) {
+  private void addRepositoryTo(project) {
```
alkamo, line 573: I had briefly moved these classes out into a helper class (and made them static), so they could be used in another spec. I later converted that spec into the OutputFormatter paradigm and no logger need the helper, so I copied them back, but I left the changes in the declarations, under the assumption that greater specificity doesn't hurt.

I later removed 'static' from `addRepositoryTo` because the `getClass` call was causing the test to fail under Java 11.

ben-manes:
Released to jcenter, will sync to the plugin portal within ~2 hours.

Thanks for popping up! Glad to see you're still around since the Flyway days when I was on a plugin writing bing.

ben-manes:
This is now out as 0.38, as the 0.37 got borked. Sorry about that.

PR #481, opened 2021-02-24
Use recommended Kotlin DSL syntax to prevent errors in IntelliJ

peterfigure:
https://docs.gradle.org/current/userguide/kotlin_dsl.html#kotdsl:containers


ben-manes:
thanks!

PR #483, opened 2021-02-26
Publish to the portal directly (fixes #473)

### LINKED ISSUE #473: JCenter to be sunset on 2020-05-01
Hi! Just a heads up- JCenter will no longer be available after 2020-05-01:
https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/

Just a note in case it makes sense to publish releases somewhere new (Maven Central, etc.)

Best,
Pete

ben-manes:
Migrates off of JCenter and publishes directly to the plugin portal. The github action is run only when a release is created.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,9 +1,8 @@
 name: deploy
 
-on:  
-  push:
-    tags:
-      - '*'
+on:
+  release:
+    types: [created]
 
 jobs:
   publish:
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy.yml
```diff
@@ -21,11 +20,26 @@ jobs:
...
       - name: Publish
         uses: burrunan/gradle-cache-action@v1.6
         with:
           remote-build-cache-proxy-enabled: false
-          arguments: clean build bintrayUpload -s --scan
+          arguments: publishPlugins -s --scan
         env:
-          BINTRAY_API_KEY: ${{ secrets.BINTRAY_API_KEY }}
-          BINTRAY_USERNAME: ${{ secrets.BINTRAY_USERNAME }}
+          GRADLE_PUBLISH_KEY: ${{ secrets.GRADLE_PUBLISH_KEY }}
```
anuraaga, line 39: Don't think these are used. Alternatively, I think you can clean up with just `-Pgradle.publish.key=${{ secrets.GRADLE_PUBLISH_KEY}} -Pgradle.publish.secret=${{ secrets.GRADLE_PUBLISH_SECRET }}` without the file mangling? GitHub scrubs secrets from all output automatically in case that's the concern.

ben-manes: oh, thanks! I used the [properties block](https://github.com/burrunan/gradle-cache-action#usage) on the action instead, as that seems a little cleaner. wdyt?

anuraaga: Ah yeah looks great

anuraaga (approved):

PR #496, opened 2021-04-08
Adding tasks getVersion and setVersion

ldenisey:
Hello,

I want to implement something like the maven goal versions:set to ease modification, mostly for CI needs. I think you plugin is right place to centralize this feature, but please do tell me if you disagree so that I publish it in a new plugin.

In term of functionality, the idea is to be able to change project current version with a single line command, either by setting a new one from scratch (like mvn versions:set) but also modifying it (like npm version).

I have created 2 tasks : getVersion mostly for dry run and setVersion that actually modify the version in gradle build or properties file.

The documentation is added in the readme, codenarc issues are fixed (except 3 that can't be) and added unit tests.

In term of design, I have created groovy and kotlin parsers in dedicated package for later reuse.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,8 @@
-
-
 # Gradle Versions Plugin
 
 In the spirit of the [Maven Versions Plugin](http://www.mojohaus.org/versions-maven-plugin),
-this plugin provides a task to determine which dependencies have updates. Additionally, the plugin
-checks for updates to Gradle itself.
+this plugin provides tasks to determine which dependencies have updates and to manage version updates.
+Additionally, the plugin checks for updates to Gradle itself.
 
 You may also wish to explore additional functionality provided by,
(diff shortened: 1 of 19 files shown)


ben-manes:
I appreciate the amount of work you did here, especially with tests. However, I think this is better managed as its own plugin. 

Adding to this plugin implies that I'd be the long term owner, whereas you're surely the best maintainer. It is also very easy for this plugin to fall into supporting opinionated workflows such as how to manage the version catalog or, here, how the build version should be maintained. Instead, I think plugins should either favor being focused and simple or an opinionated composite workflow. The latter can usually be built on top of other plugins, so they script opinionated functionality.

In this case a new plugin would be quite neat. You might also look at some other existing build versioning and release management plugins for inspiration for how you would like your own to evolve.

ldenisey:
Ok, sure. Regarding existing build versioning and release management plugins, do you have some in mind ?

PR #497, opened 2021-04-10
Bump gradle to 7.0

Goooler:


ben-manes:
Looks like it failed because the old plugin properties file is now generated by Gradle itself. You could remove it and try again.

Goooler:
How to solve this ? @ben-manes 

image

ben-manes:
```
[groovy-nio is loaded in version 3.0.7 and you are trying to load version 2.5.4
```

I bet this is because we [use](https://github.com/ben-manes/gradle-versions-plugin/blob/2cec181310712ab5085ae3b0fc1a07051e9bdc01/build.gradle#L59-L61) Spock 1.x, whereas the [upgrade guide](https://docs.gradle.org/7.0/userguide/upgrading_version_6.html#building_gradle_plugins_with_groovy_3) says it requires 2.x. You could try upgrading, as we don't do anything special in our tests.

Goooler:
> ```
> [groovy-nio is loaded in version 3.0.7 and you are trying to load version 2.5.4
> ```
> 
> I bet this is because we [use](https://github.com/ben-manes/gradle-versions-plugin/blob/2cec181310712ab5085ae3b0fc1a07051e9bdc01/build.gradle#L59-L61) Spock 1.x, whereas the [upgrade guide](https://docs.gradle.org/7.0/userguide/upgrading_version_6.html#building_gradle_plugins_with_groovy_3) says it requires 2.x. You could try upgrading, as we don't do anything special in our tests.

Not work

ben-manes:
Well, that looks like it no longer brings in junit4 as a transitive dependency.

ben-manes:
After adding `testImplementation 'junit:junit:4.13.2'` it compiles, but no tests are run.

ben-manes:
and adding `useJUnitPlatform()` to enable Spock 2's usage of the JUnit5 runner, and tests fail because they use JUnit 4 Rules for configuration. So you'd need to migrate tests to JUnit 5 entirely to move this forward.

I assume you didn't mean to take on that work and can close, and someday we'll do that lift.

ben-manes:
See the [gradle7](https://github.com/ben-manes/gradle-versions-plugin/tree/gradle7) branch, where I migrated OutputFormatterSpec so that it passes. This required replacing the `@Rule` for a tempDir and `compile` with `implementation`. When all are migrated, then the junit4 dependency can be removed.

Goooler:
Ok, let you continue, thx.

PR #503, opened 2021-04-27
Refine code

Goooler:
- Remove redundant semicolons
- Replace method invoke with operator
- Make methods static
- Declare `getRowString(Dependency item)`

### DIFF
--- a/src/main/groovy/com/github/benmanes/gradle/versions/reporter/HtmlReporter.groovy
+++ b/src/main/groovy/com/github/benmanes/gradle/versions/reporter/HtmlReporter.groovy
@@ -124,14 +124,14 @@ class HtmlReporter extends AbstractReporter {
     }
   }
 
-  private def getCurrentRows(Result result) {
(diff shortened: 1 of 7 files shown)


ben-manes:
I don't think `getRowString` can be extracted directly without passing a few other parameters. It looks like one case uses `item.getVersion()` while another has `getDisplayableVersion(item)`. I agree the code could be shared into a helper method, but that refactoring looks wrong. Can you carefully check the code and compare an html report after the fixes?

Goooler:
> I don't think `getRowString` can be extracted directly without passing a few other parameters. It looks like one case uses `item.getVersion()` while another has `getDisplayableVersion(item)`, corresponding to different html formats. I agree the code could be shared into a helper method, but that refactoring looks wrong. Can you carefully check the code and compare an html report after the fixes?

Sorry, had not noticed that, reverted now.

ben-manes:
Thanks!

PR #504, opened 2021-04-29
Fix conflict locally project plugin vs global defined

sshuvalov:
`Failed to apply plugin 'com.github.ben-manes.versions'.
 > Cannot add task 'dependencyUpdates' as a task with that name already exists.`
 when global added plugin via https://github.com/ben-manes/gradle-versions-plugin#using-a-gradle-init-script


ben-manes:
Thanks!

Issue #508, opened 2021-05-12
Refactoring Request: HtmlReporter#getDisplayableVersion(...)


mroeppis:
As I wrote in #507 I used an included build of the versions plugin for debugging purposes. When I set `outputFormatter` to `html` and ran the `dependencyUpdates` Task it failed with a strange message. Something like:

```groovy
HtmlReporter cannot be cast to an instance of HtmlReporter.
```

While this was rather unspecific the stacktrace pointed to the following line:

https://github.com/ben-manes/gradle-versions-plugin/blob/6f7b95d52cbbaa9023384b9c6fb84c54937f712a/src/main/groovy/com/github/benmanes/gradle/versions/reporter/HtmlReporter.groovy#L198

As soon as I replaced `getDisplayableVersion(item)` by a static string things went fine. And although this shouldn't happen I somehow have the feeling that this call tried to delegate to `getDisplayableVersion(DependencyOutdated item)` which actually can't be accessed from a static context.

Anyway, this issue should not deal with the error I experienced. I can't even reproduce it when not running the plugin as an included build. The whole point is about the method itself. At the moment we have these two versions:
* https://github.com/ben-manes/gradle-versions-plugin/blob/6f7b95d52cbbaa9023384b9c6fb84c54937f712a/src/main/groovy/com/github/benmanes/gradle/versions/reporter/HtmlReporter.groovy#L205
* https://github.com/ben-manes/gradle-versions-plugin/blob/6f7b95d52cbbaa9023384b9c6fb84c54937f712a/src/main/groovy/com/github/benmanes/gradle/versions/reporter/HtmlReporter.groovy#L216

There are 4 occasions when `getDisplayableVersion(...)` is called. It happens in the static `get...Rows( Result result)` methods. All of them would use number 2 from the list above. So this raises the question why number 1 actually exists?

Though the plugin does not fail because of this at the moment I'd like to encourage you to refactor the two methods (e.g. rename, merge, ...) to ensure correct usage. Using a parameter type in 1 that extends that in 2 looks dangerous if you ask me.

---
I'm using version `0.38.0` of the plugin running with Gradle `6.7.1`

ben-manes:
I agree. This was a contribution and wasn't spotted in the review. A PR to make it cleaner would be appreciated.

PR #512, opened 2021-05-15
Issue #508: Refactoring Request: HtmlReporter#getDisplayableVersion(...)

mroeppis:
**Changes Applied**
* dissolved getDisplayableVersion(Dependency) and moved implementation to callers since this is shorter
* changed signature of getDisplayableVersion(DependencyOutdated) to getDisplayableVersion(VersionAvailable) since that parameter type is actually needed in implementation

**Status Quo**
Generating the version string is now done differently for DependencyOutdated than for other Dependency types (...as originally intended, I guess).


ben-manes (approved):
Thanks!

Issue #509, opened 2021-05-13
Report Shows Same Dependencies With Different Actual Version in Current and Outdated Section


mroeppis:
Using the default plugin configuration this is what shows up in my report:

```text
The following dependencies are using the latest milestone version:
 - ch.qos.logback:logback-classic:none
 - ch.qos.logback:logback-core:none
 - com.fasterxml.jackson.core:jackson-annotations:none
 - ...

The following dependencies have later milestone versions:
 - ch.qos.logback:logback-classic [1.2.3 -> 1.3.0-alpha5]
     http://logback.qos.ch
 - ch.qos.logback:logback-core [1.2.3 -> 1.3.0-alpha5]
     http://logback.qos.ch
 - com.fasterxml.jackson.core:jackson-annotations [2.12.2 -> 2.12.3]
     http://github.com/FasterXML/jackson
 - ...
```

I'm using a BOM platform module with exactly one version for each dependency. That's why I wonder how the report can introduce such ambiguity.

Is there a special meaning about the `none` versioned dependencies? If not it would make the report easier to understand if it didn't have those. One way of excluding them would be in:

https://github.com/ben-manes/gradle-versions-plugin/blob/6f7b95d52cbbaa9023384b9c6fb84c54937f712a/src/main/groovy/com/github/benmanes/gradle/versions/updates/DependencyUpdatesReporter.groovy#L225-L228

where a custom filter could be applied to the passed `dependencies`.

---

I'm using version `0.38.0` of the plugin running with Gradle `6.7.1`

ben-manes:
I believe `none` is used to pass it through when the version is absent, as the `version` string is required when resolving the query. This was prior to Gradle's BOM support where this occurred for file-based dependencies, e.g. `libs/foo.jar`. Perhaps now an empty version is supported by Gradle, but I have not tested it. (This plugin was originally written for Gradle 1.0)

I don't recall the code, but maybe the latest section could display the resolved version instead of the original. They would otherwise be the same, except for BOMs where it would show you what you are using. We could also strip off the "none" for file-based dependencies as it has no meaning for the user.

P.S. PRs are welcome! 🙂

mroeppis:
If different Gradle versions treat an empty version string differently at resolution then it's probably not safe to remove the `none` keyword in general. But removing it from the report seems safe to me. That's why I'd prefer to stick with my idea from above.

@ben-manes : If you still think there is another solution then please go into details.

[...]

Issue #507, opened 2021-05-12
Report Misses Dependencies because of Unchecked Configurations


mroeppis:
I'm missing dependencies when checking the report. When debugging the plugin using an included build I discovered that the configurations holding the missing dependencies are being skipped because they are not classified "useful":

https://github.com/ben-manes/gradle-versions-plugin/blob/6f7b95d52cbbaa9023384b9c6fb84c54937f712a/src/main/groovy/com/github/benmanes/gradle/versions/updates/DependencyUpdates.groovy#L70-L78

When running the above code with `resolve(...)` for all configs more dependencies are reported.

Since the reason behind the `isUsefulConfiguration` flag is not self-explanatory it would be nice to know why it's necessary. However, there should be another way to manage this.

I also found out that `Failed to resolve ...` error messages may occur because some constraints inside selected configurations can't be copied. It happens when trying to copy the configuration itself:

https://github.com/ben-manes/gradle-versions-plugin/blob/6f7b95d52cbbaa9023384b9c6fb84c54937f712a/src/main/groovy/com/github/benmanes/gradle/versions/updates/Resolver.groovy#L273

Since this is a Gradle issue I [raised](https://github.com/gradle/gradle/issues/17179) one on their end.

---
I use the latest plugin version `0.38.0` running with Gradle `6.7.1`

ben-manes:
I believe the problem that we had is the whole `canBeResolved` and `canBeConsumed` mess when Gradle changed configurations to not be resolvable. However that isn't properly set and, if I recall correctly, some backdoors allow some to bypass it. As this only applied to later versions Gradle, it was also a compatibility pain.

I am fine if we can refine this or somehow to it correctly. You can see the discussion in https://github.com/ben-manes/gradle-versions-plugin/pull/378 and the linked issues.

mroeppis:
What would happen if `isUsefulConfiguration` was removed? Is this a Kotlin issue? If so, it might be necessary for some people to restrict configurations as is done right now. On the other hand it would only be fair to allow more configurations to be checked if wanted by a user.

Could we define an input closure allowing to configure configurations?

ben-manes:
[...]

PR #522, opened 2021-05-26
Issue #509: Report Shows Same Dependencies With Different Actual Version in Current and Outdated Section; Issue #507: Report Misses Dependencies because of Unchecked Configurations

mroeppis:
**Issue #509: Report Shows Same Dependencies With Different Actual Version in Current and Outdated Section**
* implemented a filter for "none" versioned dependencies -> if there is another dependency with same group and artifact ID that has a concrete version then the none versioned dependency will be discarded -> "none" versions that pass the filter are assumed to have been declared without version by the user and must therefore be categorized "undeclared"
* added new report category "undeclared" -> dependencies that were declared without version are no longer categorized up-to-date
* added the new "undeclared" category to several test cases
* some new test resources had to be added for dependency resolution to succeed -> otherwise those would have been categorized "unresolved"
* I have thought long about the test case "Single project with flatDir repository" since this is about file dependencies. The thing is we assign a query dependency version "none" in this case which is why a latest version is not retrieved by Gradle. That's why we cannot tell if such a dependency is truly up-to-date. As for now such dependencies will be categorized "undeclared". Someday we might find a way to map file dependencies to G:A:V coordinates to perform a regular version check but that's out of scope right now
* there is one _limitation_, though: If a user declares a dependency without version and again (maybe in another project) with version then the one with version wins and the "none" versioned dependency won't be reported (...although it was explicitly declared by the user) -> however, a manual lookup of dependency occurences will be necessary anyway, so the loss of information should be minimal

**Issue #507: Report Misses Dependencies because of Unchecked Configurations**
* removed "isUsefulConfiguration" check -> from now on we analyze all configurations
* configuration resolution errors are logged at "info" -> this is motivated by the fact that Gradle itself triggers the error, so we simply document that behavior


ben-manes (approved):
This looks really good, thanks @mroeppis!

I tested it in our `examples/groovy` build, which incorrectly configured BOM dependencies. Those lack a version, so your new section comes out nicely. If fixed to `bom platform(...spring...)` then it resolves the implicit version and shows an upgrade, as intended.

You can merge and release this now that you have commit access. To release, simply update the version in [build.gradle](https://github.com/ben-manes/gradle-versions-plugin/blob/9bef5cb55bee63f2e51b3456c1172102cf4fb593/build.gradle#L14) and perform a github release. This will automatically [publish](https://github.com/ben-manes/gradle-versions-plugin/blob/master/.github/workflows/deploy.yml) to the plugin portal.

mroeppis:
Thank you. I was wondering by the way when you could cut a new release. So it sounds good to read that.
I'd like to merge this but the page still says:

> Only those with write access to this repository can merge pull requests.

Could you please verify again?

ben-manes:
It looks like you have not accepted the invitation yet. You should be able to visit the following link to accept.
https://github.com/ben-manes/gradle-versions-plugin/invitations

Screen Shot 2021-05-26 at 5 45 36 PM

mroeppis:
Thanks. I must have overlooked that.

PR merged.

PR #532, opened 2021-06-29
Issue #528: also check 'strictly' dependency constraints

### LINKED ISSUE #528: Checking dependency constraints not working
Although there is an option to enable checking the dependency constrains for updates, it doesn't work.

Example:
```
plugins {
    id 'java'
    id 'com.github.ben-manes.versions' version "0.39.0"
}
group = 'com.example'
sourceCompatibility = '11'
repositories { mavenCentral() }
dependencyUpdates.checkConstraints = true

dependencies {
    implementation("ch.qos.logback:logback-classic:1.2.3")
    constraints {
        constraints {
            implementation('org.slf4j:slf4j-api') { version { strictly '1.7.24' } }
        }
    }
}
```

Running the dependencyUpdates task results in the following output:
```
 - com.github.ben-manes.versions:com.github.ben-manes.versions.gradle.plugin:0.39.0
 - org.slf4j:slf4j-api:1.7.24

 - ch.qos.logback:logback-classic [1.2.3 -> 1.3.0-alpha5]
     http://logback.qos.ch
```

Checking on [maven central](https://mvnrepository.com/artifact/org.slf4j/slf4j-api) shows, that the newest version of slf4j-api is in fact currently 2.0.0-alpha1.

Looking at the Resolver.groovy, around Line 154:
```groovy
    copy.dependencies.clear()
    copy.dependencies.addAll(latest)
    copy.dependencies.addAll(inherited)
```

This part cleans up the dependencies from the copied configuration prior running the resolver and doing the actual version checks.

Clearing the dependencyConstraints at that point makes it work like expected for me:
```groovy
    copy.dependencies.clear()
    copy.dependencies.addAll(latest)
    copy.dependencies.addAll(inherited)
    copy.dependencyConstraints.clear()
```

I think the issue here might be the `strictly` constraint, as it effectively disables any further resolution. As we need it to be strictly in our project to downgrade some transitive dependencies, we may not change it to another value, like `require` or `prefer`.

Could you please consider removing all constraints prior resolving the dependencies?

Thanks,
Simon

bratkartoffel:
This PR removes the dependency constraints, enabling update checks for 'strictly' constraints.
Closes #528

### DIFF
--- a/src/main/groovy/com/github/benmanes/gradle/versions/updates/Resolver.groovy
+++ b/src/main/groovy/com/github/benmanes/gradle/versions/updates/Resolver.groovy
@@ -155,6 +155,9 @@ class Resolver {
     copy.dependencies.addAll(latest)
     copy.dependencies.addAll(inherited)
 
+    // Remove all dependency constraints, even the 'strictly' ones
+    copy.dependencyConstraints.clear()
+
     addRevisionFilter(copy, revision)
     addAttributes(copy, configuration)
     addCustomResolutionStrategy(copy, currentCoordinates)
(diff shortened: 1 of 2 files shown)


ben-manes:
@anuraaga can you please review this change as I am not very familiar with constraints.

ben-manes:
Closing per discussion on #528 where we decided to not merge this.

PR #543, opened 2021-07-13
Update github action versions

Goooler:
- Bump `setup-java` to v2
- Replace minor versions with major versions (will auto keep the latest)


ben-manes:
thanks! The error indicates that the syntax is 8 rather than 1.8

Goooler:
> thanks! The error indicates that the syntax is now 8 rather than 1.8

Modified.

PR #545, opened 2021-07-13
Refine code

Goooler:


ben-manes:
Sorry, but I disagree with these changes. 😢
 
1. The full import literals is generally a better practice than a wildcard
2. `reportfileName ` is unfortunately a public field that is part of the configuration dsl. A fixed name would need to be backwards compatible.
3. method params should be indented
4. The `@Override` is a good practice, but a nit.

PR #556, opened 2021-08-20
Bump gradle to 6.9.1

Goooler:


ben-manes:
Hi @Goooler,

I appreciate the help, trying to keep everything up to date, and polished. However it is also a bit noisy since otherwise development is fairly irregular. I can give you access to push directly if you want, but I'd also be inclined to not update so frequently since it doesn't appear to have much of a positive impact.

regardless, though, thanks!

Goooler:
> Hi @Goooler,
> 
> I appreciate the help, trying to keep everything up to date, and polished. However it is also a bit noisy since otherwise development is fairly irregular. I can give you access to push directly if you want, but I'd also be inclined to not update so frequently since it doesn't appear to have much of a positive impact.
> 
> regardless, though, thanks!

Thank u.

PR #568, opened 2021-11-12
Remove Bintray links from default HTML reporting

ed-george:
As widely publicised, [Bintray was sunset in May 2021](https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/). 

The Bintray links generated by the default HTML report no longer go to the correct location and, as such, should be removed 👍


ben-manes (approved):
Thanks!

Issue #576, opened 2021-12-22
Gradle 7.3.2 added a constraint and it show as a depencendy update


ErwanLeroux:
Hello, I noticed something when upgrading from gradle 7.3 to 7.3.2, log4j-core is shown as an update even if the dependency is not in the project

```
The following dependencies have later milestone versions:
 - org.apache.logging.log4j:log4j-core [2.16.0 -> 2.17.0]
     https://logging.apache.org/log4j/2.x/
```

It took me a while to understand that it was because Gradle added a constraint in 7.3.2, to help mitigate https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-45105
Mitigations for log4j vulnerability in Gradle builds (logj4:2.16.0) https://github.com/gradle/gradle/issues/19300

Should the plugin tell us about this constraint if we can't do anything about it ? And if no, how can we configure the plugin to ignore that particular constraint, like we can do with 
Could the report be more explicit that the update is from a constraint instead a dependency ?

I made a project to reproduce the context : https://github.com/ErwanLeroux/dU-mcve but it can resumed as 
```
plugins {
    java
    id("com.github.ben-manes.versions") version "0.39.0"
}

tasks.withType<com.github.benmanes.gradle.versions.updates.DependencyUpdatesTask> {
    checkConstraints = true
}
```

I look forward to ear from you, I really like this plugin :)

ben-manes:
The `checkConstraints` feature was contributed by @anuraaga, so he might have recommendations on how we should handle this case. The comment indicates that not matching is desired for transitive scenarios in order to detect these type of problems. Of course he wrote that assuming the build author added that intentionally, whereas Gradle's own is surprising noise. I think to differentiate we'd have to resolve the dependency graph to see if it was listed transitively. It could be confusing regardless of what we do, so I am not sure about the best approach here.

https://github.com/ben-manes/gradle-versions-plugin/blob/a51972baeeffc8db8e86c066a40c0a5db9991887/src/main/groovy/com/github/benmanes/gradle/versions/updates/Resolver.groovy#L292-L301

anuraaga:
From what I can tell Gradle adds the constraint to the `buildscript` dependencies, not the project dependencies. 

https://github.com/gradle/gradle/commit/6aeddcf0c146ea195380314026a0fdc418cb8d28

[...]

PR #577, opened 2021-12-27
Exclude constraints check from buildSrc (#576)

ErwanLeroux:
I applied some suggestions from @ben-manes and @anuraaga to try to fix the issue added in 7.3.2 and 6.9.2
Your input is very welcome :)


src/main/groovy/com/github/benmanes/gradle/versions/updates/Resolver.groovy
```diff
@@ -416,7 +418,7 @@ class Resolver {
   }
 
   private boolean supportsConstraints(Configuration configuration) {
```
ben-manes, line 420: Can you add a little JavaDoc explaining the `fromBuildSrc` condition so that the rational is easier to recall for future readers?

ben-manes:
This lgtm, thanks! I'll leave it to @anuraaga to approve for merging and then we can cut a new release for you.

src/test/groovy/com/github/benmanes/gradle/versions/ConstraintsSpec.groovy
```diff
@@ -131,4 +131,70 @@ final class ConstraintsSpec extends Specification {
     result.output.contains('No dependencies found.')
     result.task(':dependencyUpdates').outcome == SUCCESS
   }
+
+  def "Does not show updates for an gradle constraint added in 7.3.2"() {
```
anuraaga, line 135: Can you add a test showing the update for log4j if defined in the build file for normal project dependencies?

src/test/groovy/com/github/benmanes/gradle/versions/ConstraintsSpec.groovy
```diff
@@ -131,4 +131,70 @@ final class ConstraintsSpec extends Specification {
...
+        }
+
+        tasks.withType<com.github.benmanes.gradle.versions.updates.DependencyUpdatesTask> {
+            checkConstraints = true
+        }
+      """.stripIndent()
+
+    when:
+    def result = GradleRunner.create()
+      .withGradleVersion('6.9.2')
```
anuraaga, line 187: I think you can use this sort of pattern to test the multiple versions

https://github.com/ben-manes/gradle-versions-plugin/blob/master/src/test/groovy/com/github/benmanes/gradle/versions/DifferentGradleVersionsSpec.groovy#L100

src/test/groovy/com/github/benmanes/gradle/versions/ConstraintsSpec.groovy
```diff
@@ -131,4 +131,70 @@ final class ConstraintsSpec extends Specification {
...
+      """.stripIndent()
+
+    when:
+    def result = GradleRunner.create()
+      .withGradleVersion('7.3.2')
+      .withProjectDir(testProjectDir.root)
+      .withArguments('dependencyUpdates')
+      .forwardStdError(srdErrWriter)
+      .withPluginClasspath()
+    //.forwardOutput() // Affiche les logs dans la console
```
anuraaga, line 159: Can you remove these debug lines?

src/main/groovy/com/github/benmanes/gradle/versions/updates/Resolver.groovy
ErwanLeroux: I will definitely add some javadoc and comment ^^
Do you think fromBuildSrc should be a parameter like checkConstraints ?

src/test/groovy/com/github/benmanes/gradle/versions/ConstraintsSpec.groovy
ErwanLeroux: Right, I thought about it at some point, and forgot :)

src/main/groovy/com/github/benmanes/gradle/versions/updates/Resolver.groovy
ErwanLeroux: Or rather, should there be a parameter like ignoreConstraintsFromBuildSrc at the same level as checkConstraints ?

src/test/groovy/com/github/benmanes/gradle/versions/ConstraintsSpec.groovy
ErwanLeroux: done

src/test/groovy/com/github/benmanes/gradle/versions/ConstraintsSpec.groovy
ErwanLeroux: done

src/test/groovy/com/github/benmanes/gradle/versions/ConstraintsSpec.groovy
ErwanLeroux: done

src/main/groovy/com/github/benmanes/gradle/versions/updates/Resolver.groovy
ben-manes: If we did that, I’d do it as an enablement flag. Then you just pass in a check constraints boolean for each, e.g. checkBuildEnvironmentConstraints, and you don’t need a suppression flag. How about that?

anuraaga (approved):
Thanks!

src/main/groovy/com/github/benmanes/gradle/versions/updates/Resolver.groovy
```diff
@@ -59,14 +59,16 @@ class Resolver {
   final Project project
   final Action<? super ResolutionStrategyWithCurrent> resolutionStrategy
   final boolean checkConstraints
+  final boolean fromBuildSrc
   final ConcurrentMap<ModuleVersionIdentifier, ProjectUrl> projectUrls
 
   Resolver(Project project, Action<? super ResolutionStrategyWithCurrent> resolutionStrategy,
-           boolean checkConstraints) {
+           boolean checkConstraints, boolean fromBuildSrc) {
```
ben-manes, line 66: The caller could just pass in false for checkConstraints if from buildsrc. Then you don’t need any changes to this class 🙂

src/main/groovy/com/github/benmanes/gradle/versions/updates/Resolver.groovy
```diff
@@ -131,7 +137,7 @@ class Resolver {
       copy.setCanBeResolved(true)
     }
 
-    // Resolve using the latest version of explicitly declaired dependencies and retains Kotlin's
+    // Resolve using the latest version of explicitly declared dependencies and retains Kotlin's
```
ben-manes, line 140: Thanks!

src/main/groovy/com/github/benmanes/gradle/versions/updates/Resolver.groovy
ErwanLeroux: Hum, I'm not sure what you meant. Something like this ?
```
tasks.withType<DependencyUpdatesTask> {
  checkProjectsConstraints = true
  checkBuildEnvironmentConstraints = true
}
```

src/main/groovy/com/github/benmanes/gradle/versions/updates/Resolver.groovy
ErwanLeroux: nice catch, it's done :)

src/main/groovy/com/github/benmanes/gradle/versions/updates/Resolver.groovy
ben-manes: Yes. You don't need to pass `fromBuildSrc` into `Resolver` as it only needs `checkConstraints`, which can be determined at the task level. That configuration example is simpler to reason about and then is passing the param value through from the task to the resolver. For backwards compatibility I'd keep `checkProjectsConstraints` called `checkConstraints` or keep it as an alias it if you prefer the clearer terminology.

ben-manes: (Also this is optional based on if you prefer it or now. I'm fine merging as is or having this new configuration option)

ErwanLeroux: I would rather let the user have the option to check those constraints, not all of them will come from gradle.

ErwanLeroux: I added the configuration option, it feel way better than just ignoring those external constraints :)
I had to adapt the tests, since a new log4j vulnerability was patched.
There is something bothering about the test, I think I should not have seen an update to log4j-core:2.17.1 since i specified only 2.16.0 and 2.17.0 in the test arborescence, so I must be missing something

build.gradle
```diff
@@ -11,7 +11,7 @@ repositories {
 }
 
 group = 'com.github.ben-manes'
-version = '0.39.0'
+version = '0.39.1-SNAPSHOT'
```
anuraaga, line 14: Just noticed this - should we revert this since I guess it's for debugging?

src/main/groovy/com/github/benmanes/gradle/versions/updates/Resolver.groovy
anuraaga: FWIU the tests are intended to run against the mock m2 repository but I wonder if the mock is only applied to the project dependencies and not the build ones? I guess so since it's a hard problem - Gradle can easily add more build dependencies in tbe future and maintaining a mock m2 for Gradle seems challenging. I personally think the current assertions, checking that an update was presented, but not the actual presented version, make sense.

build.gradle
ben-manes: yeah, this can be incremented to `0.40.0` for the merge and release.

src/main/groovy/com/github/benmanes/gradle/versions/updates/DependencyUpdates.groovy
```diff
@@ -46,26 +46,36 @@ class DependencyUpdates {
...
       [[group: it.coordinate.groupId, name: it.coordinate.artifactId]: it.projectUrl]
     }
     return createReporter(versions, unresolved, projectUrls)
   }
 
-  private Set<DependencyStatus> resolveProjects(Map<Project, Set<Configuration>> projectConfigs) {
+  private Set<DependencyStatus> resolveProjects(Map<Project, Set<Configuration>> projectConfigs, boolean checkConstraints) {
     Set<DependencyStatus> resultStatusSet = []
     for(Project currentProject : projectConfigs.keySet()) {
+      // We ignore constraints from buildSrc configurations since some are added by gradle (https://github.com/ben-manes/gradle-versions-plugin/issues/576)
```
ben-manes, line 78: I think you can drop this line now

ben-manes:
I really like how this worked out, thank you @ErwanLeroux! When you are ready please merge and release. 🙂 

Simply set the next version in the build and creating a github release with the appropriate tag name and description. A github action will then publish it to the plugin portal, and you can use it in your project immediately afterwards.

src/main/groovy/com/github/benmanes/gradle/versions/updates/DependencyUpdates.groovy
ErwanLeroux: good catch

src/main/groovy/com/github/benmanes/gradle/versions/updates/Resolver.groovy
ErwanLeroux: you're welcome :)

ErwanLeroux:
I can't create a release without a tag, and I don't see an option to create a tag, it's probably not available to my profil.

I really liked working on the plugin (ie outside of work), it was a fun way to learn about gradle, plugin, and a bit of kotlin.  thank you @ben-manes and @anuraaga

ErwanLeroux:
And of course I understood how to create a new tag right after saying I didn't know how :D (That must be a law of development : find the answer right after asking the question)

ben-manes:
haha, yes. It's automatically done as part of the release notes. It looks like the deploy action finished and it is on the portal!

ErwanLeroux:
Indeed, I was able to use it in my project :)

PR #580, opened 2022-01-12
Remove  deblibs-gradle-plugin from the list

konstantin-kiriushyn:
Hey @ben-manes. Unfortunately we decided to stop support for https://github.com/hellofresh/deblibs-gradle-plugin. 

This PR removes a reference to it, so we can avoid redirecting people to deprecated plugin which will be really hard to integrate. If at any point we will decide to update the deblibs plugin and put it back we would let you know. 

Thank you very much!


ben-manes:
Oh okay, thanks!

PR #585, opened 2022-02-04
Prevent builds from failing w/ configuration cache enabled on Gradle 7.4+

eygraber:


src/main/groovy/com/github/benmanes/gradle/versions/updates/DependencyUpdatesTask.groovy
```diff
@@ -88,6 +88,20 @@ class DependencyUpdatesTask extends DefaultTask {
...
+  }
+
+  private boolean supportsIncompatibleWithConfigurationCache() {
+    return metaClass.respondsTo(this, "notCompatibleWithConfigurationCache", String)
+  }
+
+  private void callIncompatibleWithConfigurationCache() {
+    String methodName = "notCompatibleWithConfigurationCache"
+    Object[] methodArgs = ["The gradle-versions-plugin isn't compatible with the configuration cache"]
+    metaClass.invokeMethod(this, methodName, methodArgs)
```
ben-manes, line 104: I think you can simply invoke the method since it is dynamically typed? Our other usages only guard before making the method call directly.

ben-manes (approved):
Thanks for this and the tests!

src/main/groovy/com/github/benmanes/gradle/versions/updates/DependencyUpdatesTask.groovy
eygraber: I'm not the most well versed on Groovy because that's what I thought as well, but the IDE was showing it as red. I didn't actually try running it though.

ben-manes: haha, me neither. I just use VSCode and pray =)

ben-manes:
You are welcome to merge and make a release. To do that simply increment the build's version, create the release (it will create a tag), and the github action will publish it to the portal.

src/main/groovy/com/github/benmanes/gradle/versions/updates/DependencyUpdatesTask.groovy
eygraber: I just double checked and it won't run. I'm assuming the compiler can't see it because the version of the Gradle API needs to be bumped, but not sure if that is desired at the moment.

ben-manes: Sounds good, thanks!

PR #593, opened 2022-03-25
ignore failOnDynamicVersions() to fetch latest versions

### LINKED ISSUE #592: failOnDynamicVersions() prevents fetching latest version
preventing dynamic version updates for configurations as in https://docs.gradle.org/current/userguide/resolution_strategy_tuning.html#reproducible-resolution prevents `dependencyUpdates` from fetching the latest versions of all dependencies.
```
configurations.all {
    resolutionStrategy {
        failOnDynamicVersions()
    }
}
```
is there a recommended workaround or can the (global) resolution strategy be disabled just for the configurations added by the `Resolver`? :crystal_ball:

keykey7:
Provides a way to fetch the latest version despite having `resolutionStrategy.failOnDynamicVersions()` active.
Gradle does not expose an official way to disable this check once activated currently. The proposed solution relies on the implementation of the internal `DefaultResolutionStrategy`, but is generally backwards compatible.

resolves #592

### DIFF
--- a/src/main/groovy/com/github/benmanes/gradle/versions/updates/Resolver.groovy
+++ b/src/main/groovy/com/github/benmanes/gradle/versions/updates/Resolver.groovy
@@ -41,9 +41,6 @@ import org.gradle.api.artifacts.result.ComponentArtifactsResult
 import org.gradle.api.artifacts.result.ResolvedArtifactResult
 import org.gradle.api.attributes.HasConfigurableAttributes
 import org.gradle.api.internal.artifacts.DefaultModuleVersionIdentifier
-import org.gradle.api.internal.artifacts.configurations.DefaultConfiguration
-import org.gradle.api.internal.artifacts.configurations.ResolutionStrategyInternal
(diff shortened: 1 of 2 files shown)


src/main/groovy/com/github/benmanes/gradle/versions/updates/Resolver.groovy
```diff
@@ -41,6 +41,9 @@ import org.gradle.api.artifacts.result.ComponentArtifactsResult
 import org.gradle.api.artifacts.result.ResolvedArtifactResult
 import org.gradle.api.attributes.HasConfigurableAttributes
 import org.gradle.api.internal.artifacts.DefaultModuleVersionIdentifier
+import org.gradle.api.internal.artifacts.configurations.DefaultConfiguration
+import org.gradle.api.internal.artifacts.configurations.ResolutionStrategyInternal
+import org.gradle.api.internal.artifacts.ivyservice.resolutionstrategy.DefaultResolutionStrategy
```
ben-manes, line 46: looks like you don't need these anymore?

ben-manes (approved):

ben-manes:
Thanks! great idea and approach :)

PR #600, opened 2022-04-28
add more robust gitignore

jaredsburrows:
I keep seeing `.DS_Store` from my MacBook. So I added a more robust `.gitignore` that should handle most cases.

@ben-manes


ben-manes (approved):
This looks good except that you dropped the existing ones instead of adding. If you bring those back then lgtm

jaredsburrows:
@ben-manes I merged both of the `.gitignore` files and removed the duplicates.

.gitignore
```diff
@@ -1,15 +1,96 @@
...
+# Files for the Dalvik VM
+*.dex
+
+# Java class files
+*.class
+
+# Gradle files
+.gradle/
+build/
+buildSrc/
```
eygraber, line 34: Out of curiosity, why ignore `buildSrc`? That's usually important.

jaredsburrows: https://github.com/ben-manes/gradle-versions-plugin/pull/605.

PR #602, opened 2022-04-28
gradle 7.4.2

jaredsburrows:
- Gradle 7.4.2
 - Removed `com.github.ben-manes.versions.properties`
```
Caused by: org.gradle.api.InvalidUserCodeException: Entry META-INF/gradle-plugins/com.github.ben-manes.versions.properties is a duplicate but no duplicate handling strategy has been set. Please refer to https://docs.gradle.org/7.4.2/dsl/org.gradle.api.tasks.Copy.html#org.gradle.api.tasks.Copy:duplicatesStrategy for details.
```

@ben-manes


jaredsburrows:
@ben-manes not sure what to do about 
```
> Receiver class com.github.benmanes.gradle.versions.updates.Coordinate does not define or inherit an implementation of the resolved method 'abstract java.lang.Object getProperty(java.lang.String)' of interface groovy.lang.GroovyObject.
```

ben-manes:
I'm not sure. The simplest might be to make it a Java class as it is just a simple data holder.

jaredsburrows:
@ben-manes Are you open to more pure Java sources in this repo? Maybe long term, we can rewrite to Kotlin :)

ben-manes:
yeah, that's fine. I don't know Groovy beyond just figuring out how to write the plugins back when that was Gradle's preferred option. I'm not opinionated, with the only minor caveat that I haven't worked with kotlin but that makes it mostly equivalent to my Groovy expertise.

jaredsburrows:
@ben-manes Finally this is now passing :)

ben-manes (approved):

PR #604, opened 2022-04-28
remove XmlSlurper

jaredsburrows:
- Removed `XmlSlurper` since it is deprecated
  - See: https://docs.groovy-lang.org/3.0.7/html/gapi/groovy/util/XmlSlurper.html
- Format files

@ben-manes


ben-manes:
Looks like there are 7 test failures. lgtm once resolved.

PR #607, opened 2022-04-28
[WIP] Migrate to Kotlin - Part 1

jaredsburrows:
- Add temporary `api` module to convert code to Kotlin
- Can't mix Groovy and Kotlin together, easier to make separate module
- Keep tests in Groovy for Spock

Pushing commits here for conversion and CI

@ben-manes


ben-manes (approved):

jaredsburrows:
@ben-manes `git mv` does not need to keep the history. Is there a better way?

ben-manes:
Unfortunately my understanding is that is the best available and by design doesn't handle history elegantly. One can figure it out by using `git log --follow`, or just spelunking through prior commits / tags.

jaredsburrows:
@ben-manes I think this is enough for a first pass or "part 1"?

ben-manes:
cool! 🙂

PR #608, opened 2022-04-29
declare explicit types

jaredsburrows:
- Declare the types explicitly. This helps make the codebase more readable and easier to convert to Kotlin in the future.
 - Fix formatting

@ben-manes


ben-manes (approved):
That looks great, thank you!

PR #609, opened 2022-04-30
more clean up, use of double quotes and @Nullable annotations

jaredsburrows:
- More clean up
- Use of double quotes
- Use of `@Nullable` annotations

@ben-manes


ben-manes (approved):
purely fyi,
Double quotes in Groovy results in a [GString](https://docs.groovy-lang.org/latest/html/api/groovy/lang/GString.html) for expression evaluation. The single quotes are used to be explicit about intent and avoid compatibility surprises.

PR #611, opened 2022-05-01
update createClasspathManifest

jaredsburrows:
- Update `createClasspathManifest`

@ben-manes


ben-manes:
lgtm but I don’t know what is the cause of those failures

jaredsburrows:
@ben-manes Looks like Spocks included version of Groovy conflicts with the current Gradle version.

PR #618, opened 2022-05-01
remove JsonBuilder

jaredsburrows:
@ben-manes

### DIFF
--- a/src/main/groovy/com/github/benmanes/gradle/versions/reporter/JsonReporter.groovy
+++ b/src/main/groovy/com/github/benmanes/gradle/versions/reporter/JsonReporter.groovy
@@ -1,7 +1,9 @@
 package com.github.benmanes.gradle.versions.reporter
 
 import com.github.benmanes.gradle.versions.reporter.result.Result
-import groovy.json.JsonBuilder
+import com.squareup.moshi.JsonAdapter
+import com.squareup.moshi.Moshi
+import com.squareup.moshi.kotlin.reflect.KotlinJsonAdapterFactory
 import groovy.transform.CompileStatic
 import groovy.transform.TupleConstructor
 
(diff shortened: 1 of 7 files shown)


src/main/groovy/com/github/benmanes/gradle/versions/reporter/JsonReporter.groovy
```diff
@@ -1,20 +1,47 @@
...
   @Override
   void write(Appendable printStream, Result result) {
-    printStream.println(new JsonBuilder(result).toPrettyString().stripMargin())
+    JsonAdapter<Result2> jsonAdapter = moshi.adapter(Result2.class)
+    .indent(" ")
+    Result2 result2 = new Result2()
+    result2.count = 1
+    result2.gradle = new GradleUpdateResults()
+//    JsonAdapter<Result> jsonAdapter =
+//      moshi.adapter(Types.newParameterizedType(Result.class, SortedSet.class, com.github.benmanes.gradle.versions.reporter.result.Dependency.class))
```
ben-manes, line 40: delete debug?

ben-manes (approved):

jaredsburrows:
I pushed this up to save it temporarily. I found more generics and other types that Moshi cannot handle. Looking into this now.

jaredsburrows:
@ben-manes For some reason, changing the types changes `<dependencies>` to `<dependencies class="sorted-set">`.

jaredsburrows:
@ben-manes I was able to fix this via `xstream.aliasSystemAttribute(null, "class")`.

ben-manes:
By switching to `Set` did you lose sorted order? If so, please add a TreeSet copy. We don't want to display the results in hash order as confusing.

jaredsburrows:
@ben-manes Can you explain more here? Moshi cannot handle `SortedSet`, suggested I use `Set`. Which test was checking the order?

ben-manes:
I don’t think we have tests, but the output should show dependencies in sorted order. You can just sort at the usages though via a new TreeSet copy

PR #626, opened 2022-05-04
more kotlin

jaredsburrows:


api/src/main/kotlin/com/github/benmanes/gradle/versions/updates/gradle/GradleUpdateChecker.kt
```diff
@@ -20,7 +20,7 @@ class GradleUpdateChecker(
    * @return if the check for Gradle updates was enabled and, if so, the versions were fetched.
    */
   @get:JvmName("isEnabled")
-  val enabled: Boolean = true,
+  val enabled: Boolean,
```
ben-manes, line 23: this changes the default to false?

ben-manes (approved):
Please make sure that gradle update checks are enabled by default

jaredsburrows:
Closing this in favor of https://github.com/ben-manes/gradle-versions-plugin/pull/627.

api/src/main/kotlin/com/github/benmanes/gradle/versions/updates/gradle/GradleUpdateChecker.kt
jaredsburrows: This originally had no value - https://github.com/ben-manes/gradle-versions-plugin/blob/c2bad3d8661e203410c80afddcbb3ade4fdc772e/src/main/groovy/com/github/benmanes/gradle/versions/updates/gradle/GradleUpdateChecker.groovy#L26.

I can keep it `true`.

jaredsburrows: https://github.com/ben-manes/gradle-versions-plugin/blob/9f0260e419dc43be340bc048890eef1aaabfd71a/src/main/groovy/com/github/benmanes/gradle/versions/updates/DependencyUpdates.groovy#L49

PR #629, opened 2022-05-04
more kotlin

jaredsburrows:
- Convert `PlainTextReporter` to Kotlin

@ben-manes


api/src/main/kotlin/com/github/benmanes/gradle/versions/reporter/PlainTextReporter.kt
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  private fun writeUpgrades(printStream: Appendable, result: Result) {
+    val upgradeVersions = result.outdated.dependencies
+    if (upgradeVersions.isNotEmpty()) {
+      printStream.println()
+      printStream.println("The following dependencies have later $revision versions:")
+      for (dependency in upgradeVersions) {
+        val currentVersion = dependency.version
+//        printStream.println(" - ${label(dep)} [${currentVersion} -> ${dep.available[revision]}]")
```
ben-manes, line 89: ?

ben-manes (approved):

api/src/main/kotlin/com/github/benmanes/gradle/versions/reporter/PlainTextReporter.kt
jaredsburrows: Groovy:
```
  private void writeUpgrades(Appendable printStream, Result result) {
    Set<DependencyOutdated> upgradeVersions = result.outdated.dependencies
    if (!upgradeVersions.isEmpty()) {
      printStream.println()
      printStream.println("The following dependencies have later ${revision} versions:")
      for (DependencyOutdated dep in upgradeVersions) {
        String currentVersion = dep.version
        printStream.println(" - ${label(dep)} [${currentVersion} -> ${dep.available[revision]}]")
        if (dep.getUserReason()) {
          printStream.println("     ${dep.getUserReason()}")
        }
        if (dep.projectUrl != null) {
          printStream.println("     ${dep.projectUrl}")
        }
      }
    }
  }
```

Kotlin:
```
  private fun writeUpgrades(printStream: Appendable, result: Result) {
    val upgradeVersions = result.outdated.dependencies
    if (upgradeVersions.isNotEmpty()) {
      printStream.println()
      printStream.println("The following dependencies have later $revision versions:")
      for (dependency in upgradeVersions) {
        val currentVersion = dependency.version
//        printStream.println(" - ${label(dep)} [${currentVersion} -> ${dep.available[revision]}]")
        dependency.userReason?.let {
          printStream.println("     $it")
        }
        dependency.projectUrl?.let {
          printStream.println("     $it")
        }
      }
    }
  }
```

I am not sure where `${dep.available[revision]}` resolves to.

ben-manes: It’s a lookup into this
https://github.com/ben-manes/gradle-versions-plugin/blob/df508c6b4b9cf096958bd1157b92d208b1e342c4/api/src/main/kotlin/com/github/benmanes/gradle/versions/reporter/result/DependencyOutdated.kt#L9

jaredsburrows: Spoke offline.

Issue #603, opened 2022-04-28
Publish to Mavencentral again


jaredsburrows:
Official:
https://search.maven.org/search?q=a:gradle-versions-plugin

Mirror:
https://mvnrepository.com/artifact/com.github.ben-manes/gradle-versions-plugin

@ben-manes What do you think? We can use @vanniktech's https://github.com/vanniktech/gradle-maven-publish-plugin.

ben-manes:
Since the plugin portal became the defacto, is there a good reason? Those old versions are prior to the portal and this hasn't been an issue. The only recent concern is jitpack creating corrupted builds that some people are pulling in.

jaredsburrows:
It seems that the Gradle repo only stores the `jar` and `POM` file, see https://plugins.gradle.org/m2/com/github/ben-manes/gradle-versions-plugin/0.42.0/.

At least with Maven, we can store more artifacts. However, since this is a plugin, we do not need the `javadoc`, `report` or `sources` jars.

ben-manes:
yeah, I think we should favor removing those tasks instead of returning to Central

jaredsburrows:
I just wish there was support for plugin snaphots or maybe we can start publishing snapshots per build with `-SNAPSHOT`.

ben-manes:
I think that jitpack.io works okay for this most of the time

PR #636, opened 2022-05-06
clean up publshing

jaredsburrows:
Closes https://github.com/ben-manes/gradle-versions-plugin/issues/603

- Remove reports zip
- Remove tests jar
- Remove `mavenLocal()` - this is there by default

@ben-manes

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -115,13 +115,17 @@ allprojects {
   }
 
   tasks.withType(Test).configureEach {
-    maxParallelForks Runtime.getRuntime().availableProcessors()
-
     testLogging {
-      events "passed", "failed", "skipped"
+      exceptionFormat 'full'
+      showCauses true
+      showExceptions true
+      showStackTraces true
       showStandardStreams true
-      exceptionFormat "full"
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -139,32 +143,11 @@ task sourcesJar(type: Jar, dependsOn: classes) {
...
-  description = 'Create a zip of all reports.'
-  archiveClassifier = 'reports'
-  from reporting.baseDir
-}
-
-// Local published to ~/.m2 - mavenLocal() for testing
 publishing {
-  repositories {
-    mavenLocal()
-  }
```
ben-manes, line 160: This is useful for local builds to test via `examples`. Let's keep that for ourselves.

ben-manes (approved):

ben-manes:
Oh missed your comment, `Remove mavenLocal() - this is there by default`

build.gradle
jaredsburrows: This still works and is added by default.

I tested via:

```
gradlew build -x test -s
gradlew publishToMavenLocal -x test -s
```

PR #641, opened 2022-05-08
clean up DependencyUpdatesTask

jaredsburrows:
@ben-manes


src/main/groovy/com/github/benmanes/gradle/versions/updates/DependencyUpdatesTask.groovy
```diff
@@ -37,50 +37,44 @@ import org.gradle.util.ConfigureUtil
...
+  String reportfileName = System.properties.get("reportfileName", "report")
 
   @Input
   @Optional
   @Nullable
-  String getOutputFormatterName() {
-    return (outputFormatter instanceof String) ? ((String) outputFormatter) : null
-  }
+  String outputFormatterName = (outputFormatter instanceof String) ? ((String) outputFormatter) :
+    null
```
ben-manes, line 62: probably more readable as,
```
String outputFormatterName = (outputFormatter instanceof String) 
    ? ((String) outputFormatter) : null
``` 
or the Google way of
```
String outputFormatterName = (outputFormatter instanceof String) 
    ? ((String) outputFormatter)
    : null
```

ben-manes (approved):

PR #644, opened 2022-05-09
Java 17

jaredsburrows:
Supporting Java LTS versions: 8, 11, 17

https://www.oracle.com/java/technologies/java-se-support-roadmap.html

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,6 +20,7 @@ jobs:
         java-version:
           - 8
           - 11
+          - 17
 
     steps:
       - name: Checkout Project


ben-manes:
I think this is to be expected since Gradle does not support jdk18 yet

jaredsburrows:
@ben-manes It does: https://github.com/jaredsburrows/android-gif-example/pull/179

jaredsburrows:
The documentation does seem to be updated - https://docs.gradle.org/current/userguide/compatibility.html.

ben-manes:
yeah, other jdks may work or may not until officially supported. That is a bit annoying since it means one cannot test against EA builds and Gradle lags in their release cycle. They used to have 6 weeks releases between minor but lately it seems they are behind (7.4 was early Feb).

OleksandrShkurat:
I'd suggest adding JDK 17 instead of 18 as it is the latest LTS version.

tresat:
You should be able to merge this now, Gradle supports Java 19 as of [7.6](https://docs.gradle.org/7.6/release-notes.html).

The 6 week release cycle is something we're trying hard to get back to.  It looks like 8.2 is going to follow very close on the heels of 8.1.

ben-manes:
@jaredsburrows can you rebase and merge if all looks good?

jaredsburrows:
@ben-manes Odd, there should be a "Update Branch" button. Can you go to the repo's settings? https://github.com/jaredsburrows/cs-interview-questions/settings and look for the following "Always suggest updating pull request branches ":
Screenshot 2023-04-28 at 12 57 33 PM

This way folks can update via github.com and then I can click on "Rebase and merge" before submitting.

ben-manes:
done, good catch!

jaredsburrows:
@dependabot rebase

PR #660, opened 2022-05-10
replace groovy spread with java8 map

jaredsburrows:
@ben-manes


jaredsburrows:
Closing because of https://github.com/ben-manes/gradle-versions-plugin/pull/661

PR #661, opened 2022-05-10
base resolver

jaredsburrows:
@ben-manes


ben-manes (approved):

jaredsburrows:
I was able to convert all functions except for 2.

PR #678, opened 2022-05-16
remove reflection

jaredsburrows:
```
    /**
     * Gets the set of dependency constraints directly contained in this configuration
     * (ignoring superconfigurations).
     *
     * @return the set of dependency constraints
     *
     * @since 4.6
     */
    DependencyConstraintSet getDependencyConstraints();
```

```
    /**
     * Configures if a configuration can be resolved.
     *
     * @since 3.3
     */
    void setCanBeResolved(boolean allowed);
```

We can remove this also since it has been in Gradle since 4.6

@ben-manes

### DIFF
--- a/gradle-versions-plugin/build.gradle
+++ b/gradle-versions-plugin/build.gradle
@@ -30,6 +30,7 @@ dependencies {
 
   implementation localGroovy()
   implementation 'org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.6.21'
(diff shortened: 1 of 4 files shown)


ben-manes (approved):

ben-manes:
We can also bump our minimum version if that helps. We started on 1.0 and bumped when Gradle made an incompatible change that was not worth working around. Thankfully this plugin is just a report so not a big deal to pin to an old version if not upgrading the build system.

PR #689, opened 2022-08-31
Fixes reference link to Result class on docs

ubiratansoares:
Self-explained. 

Learned that when hitting [this link](https://github.com/ben-manes/gradle-versions-plugin/blob/master/src/main/groovy/com/github/benmanes/gradle/versions/reporter/result/Result.groovy)

image

Thanks for the great plugin!


ben-manes:
Thanks!

Issue #584, opened 2022-01-21
Question: Setup outputFormatter using Closure in Kotlin


jmpaya:
I am having problems setting up the outputFormatter property using a Close in Kotlin scripts.

This is the configuration I am using right now...

```kotlin
tasks.withType<DependencyUpdatesTask> {
    gradleReleaseChannel = "current"
    rejectVersionIf {
        isNonStable(candidate.version) && !isNonStable(currentVersion)
    }
    outputFormatter = { result: com.github.benmanes.gradle.versions.reporter.result.Result ->
        val updatable = result.outdated.dependencies
        if (!updatable.isEmpty()) {
            val writer = StringWriter()
            updatable.forEach { dep ->
                writer.write("> ${dep.group}:${dep.name}: ${dep.version} -> ${dep.available.release ?: dep.available.milestone}")
            }
            println(writer.toString())
        }
    }
}
```

... and the error I'm getting:
```
Cannot handle output formatter (com.github.benmanes.gradle.versions.reporter.result.Result) -> kotlin.Unit, unsupported type
```

I didn't find a complete example using Kotlin, any clue?

ben-manes:
I'm not very familiar with Kotlin so I can try this evening (PST). As that was added in the Groovy only days (as written for Gradle 1.0, and we didn't know Groovy), it checks if Groovy's `Closure` type when executing. It may be that we need to provide a proper setter that accepts an `Action` so that it will infer correctly and work with Kotlin. See the [task](https://github.com/ben-manes/gradle-versions-plugin/blob/master/src/main/groovy/com/github/benmanes/gradle/versions/updates/DependencyUpdatesTask.groovy) and [usage](https://github.com/ben-manes/gradle-versions-plugin/blob/d22df470960b6400d9451f615a9e9b460407ec04/src/main/groovy/com/github/benmanes/gradle/versions/updates/DependencyUpdatesReporter.groovy#L88) (PRs are welcome).

jpaya17:
Thanks @ben-manes. Give it a try when you find a gap because maybe I’m doing something wrong. I did review those source codes but didn’t make it work, so I did prefer asking here, maybe someone found the way.

jpaya17:
Thanks a lot @ben-manes. This is definitely what I was looking for. I'll give it a try to the enhancement that you suggested and try to raise a PR.

Issue #673, opened 2022-05-15
Kotlin DSL error


jaredsburrows:
I am still seeing the following after the recent Kotlin migration.

```
$ cd examples/kotlin/ && ../../gradlew && cd -

> Configure project :
e: /Users/<>/repo/gradle-versions-plugin/examples/kotlin/build.gradle.kts:42:19: Type mismatch: inferred type is () -> [ERROR : Cannot infer type variable TypeVariable(_L)] but ComponentFilter was expected
e: /Users/<>/repo/gradle-versions-plugin/examples/kotlin/build.gradle.kts:43:5: Unresolved reference: candidate
e: /Users/<>/repo/gradle-versions-plugin/examples/kotlin/build.gradle.kts:47:19: Type mismatch: inferred type is () -> Boolean but ComponentFilter was expected
e: /Users/<>/repo/gradle-versions-plugin/examples/kotlin/build.gradle.kts:48:5: Unresolved reference: candidate
e: /Users/<>/repo/gradle-versions-plugin/examples/kotlin/build.gradle.kts:48:41: Unresolved reference: currentVersion
e: /Users/<>/repo/gradle-versions-plugin/examples/kotlin/build.gradle.kts:55:63: Only safe (?.) or non-null asserted (!!.) calls are allowed on a nullable receiver of type String?

FAILURE: Build failed with an exception.

* Where:
Build file '/Users/<>/repo/gradle-versions-plugin/examples/kotlin/build.gradle.kts' line: 42

* What went wrong:
Script compilation errors:

  Line 42:   rejectVersionIf {
                             ^ Type mismatch: inferred type is () -> [ERROR : Cannot infer type variable TypeVariable(_L)] but ComponentFilter was expected

  Line 43:     candidate.version.isNonStable()
               ^ Unresolved reference: candidate

  Line 47:   rejectVersionIf {
                             ^ Type mismatch: inferred type is () -> Boolean but ComponentFilter was expected

  Line 48:     candidate.version.isNonStable() && !currentVersion.isNonStable()
               ^ Unresolved reference: candidate

  Line 48:     candidate.version.isNonStable() && !currentVersion.isNonStable()
                                                   ^ Unresolved reference: currentVersion

  Line 55:         if (candidate.version.isNonStable() && !currentVersion.isNonStable()) {
                                                                         ^ Only safe (?.) or non-null asserted (!!.) calls are allowed on a nullable receiver of type String?

6 errors

* Try:
[...]

PR #690, opened 2022-09-11
Support Kotlin trailing lambda syntax for custom output formatters

vjgarciag96:
Fixes https://github.com/ben-manes/gradle-versions-plugin/issues/584.

**Support Kotlin trailing lambda syntax for custom output formatters**

* This is done by offering a new API in which an output formatter can be specified as org.gradle.api.Action<Result>. Gradle automatically maps Groovy closures and Kotlin lambdas to arguments of that type (https://docs.gradle.org/current/userguide/kotlin_dsl.html#groovy_closures_from_kotlin).

* Internally, there have been some changes done to support this:
  - The type of the property that is used to set an output formatter is `Any`, and this isn't the nicer because tooling can't really hint users of the API with the correct types, and we also have to deal manually with the types of input.
  - The new type we wanted to support (Action<Result>) would run into type erasure problems if we used this existing property and we would need to have code casting the types in an unsafe way (as we can't know if a provided argument is Action<Result> or Action of something else).
  - To sort this out, we have kept the existing property for compatibility reasons (it can still be used to set String, Reporter or Closure output formatters) but we have added a new function to set Action<Result> output formatters. With this we achieve maintaining public API compatibility whilst adding support for the new Action<Result> output formatters with improved type support (internally, safe use of types, and externally, type hinting in tooling and clear public API signature).
  

**Testing**

*Manual*
Manually tested that we can now use the trailing lamdba syntax in Kotlin, and that everything works as before both in Groovy and Kotlin. Also verified that we get type hints in the IDE when using the new API:

image

*Automated*
Added a couple of tests covering the new changes.


ben-manes:
Thanks!

I believe there was still some remaining problems to work out after the kotlin migration (https://github.com/ben-manes/gradle-versions-plugin/issues/673), so we may not be in a stable state to be able to perform a release. That issue has an answer but I don't think it was followed up on.

Looking at the plugins that use ours, the [gradle-libraries-plugin](https://github.com/fkorotkov/gradle-libraries-plugin/blob/92311acf0cdd7019f7ba7fb55d7c70e98acde9ec/src/main/java/com/fkorotkov/gradle/libraries/tasks/LibrariesBaseTask.java#L44-L50) appears to call some the newly marked `internal` classes (@fkorotkov). It might be kinder to keep them public for the usages to adapt, but otherwise not try to maintain backwards compatibility.

The changes make sense and the code looks good to me. @jaredsburrows can you please review?

vjgarciag96:
@ben-manes Thanks for feedback!

> I believe there was still some remaining problems to work out after the kotlin migration (https://github.com/ben-manes/gradle-versions-plugin/issues/673), so we may not be in a stable state to be able to perform a release. That issue has an answer but I don't think it was followed up on.

Ohhh I see, thanks for the heads-up! I might have a look into those other problems too then.

> Looking at the plugins that use ours, the [gradle-libraries-plugin](https://github.com/fkorotkov/gradle-libraries-plugin/blob/92311acf0cdd7019f7ba7fb55d7c70e98acde9ec/src/main/java/com/fkorotkov/gradle/libraries/tasks/LibrariesBaseTask.java#L44-L50) appears to call some the newly marked internal classes (@fkorotkov). It might be kinder to keep them public for the usages to adapt, but otherwise not try to maintain backwards compatibility.

That makes sense to me, everything is back to public again!

jaredsburrows:
I wonder if converting from `Closure` to `Action` would help resolve the problem we are seeing in #673.

gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/DependencyUpdatesTask.kt
```diff
@@ -12,6 +14,7 @@ import org.gradle.api.tasks.Internal
 import org.gradle.api.tasks.Optional
 import org.gradle.api.tasks.TaskAction
 import org.gradle.util.ConfigureUtil
+import java.lang.IllegalArgumentException
```
jaredsburrows, line 17: This import should not be needed in Kotlin.

gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/OutputFormatterArgument.kt
```diff
@@ -0,0 +1,33 @@
...
+
+  /**
+   * An [Action] to provide a custom output formatting. Enables the use of the trailing closure/lambda
+   * syntax for output formatting definition.
+   */
+  class CustomAction(val action: Action<Result>) : OutputFormatterArgument
+
+  companion object {
+    @JvmField
+    val DEFAULT = BuiltIn(formatterNames = "text")
```
jaredsburrows, line 31: Does this have to be a `@JvmField`?

jaredsburrows (approved):

vjgarciag96:
> I wonder if converting from Closure to Action would help resolve the problem we are seeing in https://github.com/ben-manes/gradle-versions-plugin/issues/673.

If the problem there has to do with `rejectVersionIf`, `Action` can't be used because it doesn't have a return type and `rejectVersionIf`'s argument needs to return a boolen. But I think the suggestion from the comment should work (make the `ComponentFilter` interface functional). I can give it a go and put up a PR if it works.

gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/DependencyUpdatesTask.kt
vjgarciag96: That's right, removed!

gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/OutputFormatterArgument.kt
vjgarciag96: I added it to be able to access this from Groovy code in tests. But I have removed it now as it is not necessary in production code.

ben-manes (approved):
Thanks for contributing! please merge when you’re ready.

jaredsburrows:
Thanks!

PR #691, opened 2022-09-11
Fix Kotlin example problems

vjgarciag96:
Fixes https://github.com/ben-manes/gradle-versions-plugin/issues/673

It looks like the two problems with the Kotlin DSL example are fixed with the next (as suggested in the issue):

**Make ComponentFilter interface functional**

This enables the use of Kotlin's trailing lambda syntax in the `rejectVersionIf` API.

**Make ComponentSelectionWithCurrent.currentVersion not null**

* ComponentSelectionWithCurrent objects are only created in one place and the currentVersion argument passed there is never null, so it's safe to make the property non-null as it effectively never is.
* This also fixes some issues in the Kotlin example, in which the dependencyUpdates task configuration was assumming the `currentVersion` property was not-null.

**Testing**

*Manual*
I have checked that the Kotlin DSL example works after the changes.

### DIFF
--- a/gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/resolutionstrategy/ComponentSelectionWithCurrent.kt
+++ b/gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/resolutionstrategy/ComponentSelectionWithCurrent.kt
@@ -4,7 +4,7 @@ import org.gradle.api.artifacts.ComponentSelection
 
 class ComponentSelectionWithCurrent(
   private val delegate: ComponentSelection,
-  val currentVersion: String? = null,
+  val currentVersion: String,
 ) : ComponentSelection by delegate {
 
   override fun toString(): String {
--- a/gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/resolutionstrategy/ComponentFilter.kt
+++ b/gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/resolutionstrategy/ComponentFilter.kt
@@ -3,6 +3,6 @@ package com.github.benmanes.gradle.versions.updates.resolutionstrategy
 import org.gradle.api.HasImplicitReceiver
 
 @HasImplicitReceiver
-interface ComponentFilter {
+fun interface ComponentFilter {
   fun reject(candidate: ComponentSelectionWithCurrent): Boolean
 }


gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/resolutionstrategy/ComponentSelectionWithCurrent.kt
```diff
@@ -4,7 +4,7 @@ import org.gradle.api.artifacts.ComponentSelection
 
 class ComponentSelectionWithCurrent(
   private val delegate: ComponentSelection,
-  val currentVersion: String? = null,
+  val currentVersion: String,
```
vjgarciag96, line 7: This looks like a safe change from the internal use of the class, but I'm not sure if there could be some external consumers relying on this being nullable for some reason.

ben-manes: Some dependency types lack versions, like files. I’d be more concerned about if that was affected than external usages of our internal classes.

vjgarciag96: Yes, those seem to be handled with a default "none" value rather than with null (https://github.com/ben-manes/gradle-versions-plugin/blob/master/gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/Coordinate.kt#L28). 

The only place where `ComponentSelectionWithCurrent` instances are created is https://github.com/ben-manes/gradle-versions-plugin/blob/master/gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/resolutionstrategy/ComponentSelectionRulesWithCurrent.kt#L84, and at that point the `current` variable can't be null (as there is `?: return null` check right before), and `Coordinate.version` is not-null (https://github.com/ben-manes/gradle-versions-plugin/blob/master/gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/Coordinate.kt#L20).

Let me know if that makes sense to you.

ben-manes (approved):
Thanks! I think we can release as desired then, unless @jaredsburrows knows of other blockers. I think his migration from Groovy was the only other major change.

If you do want a release sometime, simply bump the version and create a GitHub release. The new tag will cause an action to publish it.

jaredsburrows (approved):

jaredsburrows:
@vjgarciag96 Thanks for the fix!

@ben-manes Once this is in, we can test against our example project and potential release a new version.

PR #695, opened 2022-09-15
update gradle test versions

jaredsburrows:
@ben-manes


ben-manes (approved):
If the minimum is now 7.0, you might remove old test versions?

jaredsburrows:
We want to enforce minimum 7.0? We have a lot of tests that still use 5.0+. If we want to enforce, I can update the check here: https://github.com/ben-manes/gradle-versions-plugin/blob/master/gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/VersionsPlugin.kt#L13.

PR #696, opened 2022-09-15
Fix version name

vjgarciag96:
It was accidentally updated, so we are rolling back to the previous value.

I just realised this was accidentally merged in one of my previous PRs (https://github.com/ben-manes/gradle-versions-plugin/blob/master/gradle.properties#L2), so putting up the this rolling back just in case this can lead to use an undesired version name.


ben-manes (approved):

ben-manes:
thanks for catching that!

PR #719, opened 2023-01-14
Address inspection hints

sschuberth:
Please have a look at the individual commit messages for the details.

### DIFF
--- a/gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/Resolver.kt
+++ b/gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/Resolver.kt
@@ -92,7 +92,7 @@ class Resolver(
     currentCoordinates: Map<Coordinate.Key, Coordinate>,
   ): Configuration {
     val latest = configuration.dependencies
-      .filter { dependency -> dependency is ExternalDependency }
+      .filterIsInstance<ExternalDependency>()
       .map { dependency ->
         createQueryDependency(dependency as ModuleDependency)
       } as MutableList<Dependency>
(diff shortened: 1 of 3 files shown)


ben-manes:
lgtm, thanks! Seems there are some minor KtLint errors, so can merge once the checks pass.

sschuberth:
> lgtm, thanks! Seems there are some minor KtLint errors, so can merge once the checks pass.

Should be addressed now, incl. on more inspection fix.

BTW, very cool that this is Kotlin now 😀

ben-manes (approved):

sschuberth:
Any idea why [lines like this](https://github.com/ben-manes/gradle-versions-plugin/actions/runs/3919910946/jobs/6701212214#step:5:1285) are gone from the actual output with my first commit in this PR? Is it really correct that these lines are expected to be there?

ben-manes:
That is the `dependency.reason`. Maybe your refactoring to Coordinate no longer provides it?

gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/Resolver.kt
```diff
@@ -402,10 +401,10 @@ class Resolver(
 
   private fun getResolvableDependencies(configuration: Configuration): List<Coordinate> {
     val coordinates = configuration.dependencies
-      .filter { dependency -> dependency is ExternalDependency }
-      .map { dependency ->
+      .filterIsInstance<ExternalDependency>()
+      .mapTo(mutableListOf()) { dependency ->
         Coordinate.from(dependency)
-      } as MutableList<Coordinate>
```
ben-manes, line 408: I would guess that this uses a different factory method now, causing the constraint reason to be lost.

sschuberth: I believe it was something different, see the update.

ben-manes (approved):

Issue #718, opened 2023-01-05
Configurations will not be able to change roles in 9.0


jvandort:
In Gradle 9.0, configurations will no longer be able to change roles. However, the following code changes the `isCanBeResolved` flag of a copied configuration: https://github.com/ben-manes/gradle-versions-plugin/blob/0e5660e819ee032a93be313f21a9dcebb5aab0dd/gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/Resolver.kt#L108-L109

Further, starting in 8.0, these copied configuration will carry the same deprecation warnings that the source configurations had. This means in 8.0, this plugin will start printing deprecation warnings when attempting to add dependencies to and resolve arbitrary copied configurations. 

Finally, there are tentative plans to remove configuration copying in 9.0.

Is the copying here strictly necessary? Are there alternative approaches which can be used to avoid copying configurations and changing their flags?

ben-manes:
We copy the configurations so that we can rewrite the dependencies to use dynamic versions. We then resolve the original and our modified version, compare the versions selected, and report on this difference. [Originally](https://discuss.gradle.org/t/add-report-to-display-dependency-updates/540) this was a detached configuration, but there were too many Gradle bugs (like repository handler's deduplication logic either adding multiple or dropping repositories), so copying was the safest option. It also meant that we could honor configuration that was not always exposed, like resolution strategies.

Those flag changes were done in a best-effort fashion, as we swallow the exception if not resolvable. They did not exist when the plugin was written (Gradle 1.x) and were quirky (broken) in their first release, so had to be added for functionality. I believe that could be removed by changing to the newer resolution approach in v4 onward, e.g. see the dependency-analyze plugin's [logic](https://github.com/dependency-check/dependency-check-gradle/blob/7849abded305238e54fe160fb9c51195a0f24bf7/src/main/groovy/org/owasp/dependencycheck/gradle/tasks/AbstractAnalyze.groovy#L396-L415).

This plugin is mostly being casually maintained by contributors who graciously offer their time. It is ten years old, for what was meant to be a stop-gap that the Gradle team would fill when they had time (and often they said as much, but deferred that work due to this plugin being available). If you would like to see major improvements then PRs are very welcome (or advocate for a native feature that replaces this plugin). Otherwise it is whenever someone in the community kindly offers their energy, so we can't specify a timeline on when this might be fixed.

jvandort:
[...]

PR #721, opened 2023-01-18
Only resolve resolvable configurations

jvandort:
Fixes: #718

Updates plugin to only resolve resolvable configurations. As a result, configurations like implementation are not resolved individually, but transitively via any resolvable configurations which extend it. To handle this change, Resolver.kt was updated throughout to use allDependencies and allDependencyConstraints in order to ensure the resolved/unresolved dependency comparisons with the latest versions include inherited dependencies from non-resolvable configurations which were previously being resolved individually.


ben-manes:
Thanks!

ben-manes:
released, thanks @jvandort

PR #724, opened 2023-02-02
Ignore constraints of own projects ("DefaultProjectDependencyConstraint")

Brummolix:
In multimodule projects having external constraints but also constraints to own sub projects it makes no sense to check for updates of the these own subprojects (when using checkConstraints=true)

Dependening on the number of own suprojects and also the network performance of the repositories this can give a huge performance boost (for example a few minutes vs. >2 hours in a concrete project)


ben-manes:
This looks like a nice improvement to me and sounds like a reasonable change to me. @anuraaga added this feature so maybe he should review it to make sure it's a good idea, as I am a little more naive about this feature.

ben-manes:
Released in v46

PR #731, opened 2023-02-20
Do not use textual 'null' for dependencies without group (#725)

### LINKED ISSUE #725: Only locally available included builds are shown as unresolved
I have build `A` with an included build `build-logic` that is a purely local included build without a published release.
The report show this as
```plain
 - null:build-logic
The exception that is the cause of unresolved state: org.gradle.internal.resolve.ModuleVersionNotFoundException: Could not find any matches for null:build-logic:+ as no versions of null:build-logic are available.
Searched in the following locations:
  - https://repo.maven.apache.org/maven2/null/build-logic/maven-metadata.xml
Required by:
    project :
```

If such a dependency that has no version available at all is coming from an included build, then it would be nice if it would not be shown as unresolved, but probably as up-to-date or not at all.

Vampire:
Fixes #725

### DIFF
--- a/gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/reporter/PlainTextReporter.kt
+++ b/gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/reporter/PlainTextReporter.kt
@@ -191,7 +191,7 @@ class PlainTextReporter(
   companion object {
     /** Returns the dependency key as a stringified label. */
     private fun label(dependency: Dependency): String {
-      return "${dependency.group}:${dependency.name}"
+      return "${dependency.group.orEmpty()}:${dependency.name}"
     }
   }
 }
--- a/gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/Resolver.kt
+++ b/gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/Resolver.kt
@@ -69,12 +69,8 @@ class Resolver(
       val resolvedCoordinate = Coordinate.from(dependency.module.id)
       val originalCoordinate = coordinates[resolvedCoordinate.key]
       val coord = originalCoordinate ?: resolvedCoordinate
-      if (originalCoordinate == null && resolvedCoordinate.groupId != "null") {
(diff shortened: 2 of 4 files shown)


ben-manes:
Thanks! Can you fix the test failures? It might just be some minor updates, e.g.

```console
DependencyUpdatesSpec > Single project with flatDir repository FAILED
    Condition not satisfied:

    unresolved.collect { it.selector }.collectEntries { dependency -> [['group': dependency.group, 'name': dependency.name]: dependency.version] } == [['group': 'null', 'name': 'guava-18.0']: NONE_VERSION]
    |          |                       |                                                                                                           |                                            |
    |          [:guava-18.0:none]      [[group:, name:guava-18.0]:none]                                                                            false                                        none
    [:guava-18.0:none]
        at com.github.benmanes.gradle.versions.DependencyUpdatesSpec.Single project with flatDir repository_closure21(DependencyUpdatesSpec.groovy:359)
        at groovy.lang.Closure.call(Closure.java:412)
        at spock.lang.Specification.with(Specification.java:191)
        at com.github.benmanes.gradle.versions.DependencyUpdatesSpec.Single project with flatDir repository(DependencyUpdatesSpec.groovy:358)
```

Vampire:
Hm, I'm not sure about the undeclared failure, whether my changes just surfaced that there is some other bug in Resolver.kt or whether my changes are problematic.
With the current state the guice dependency is also treated as hidden.

Vampire:
Can you maybe have a look?

Vampire:
Actually if I comment out the `if (originalCoordinate == null && resolvedCoordinate.groupId != "none") {` and just keep the `else`, the tests in `DependencyUpdatesSpec` all pass.
Maybe it is not necessary and not appropriate actually as there is already `Ignore undeclared (hidden) dependencies` at an earlier place in that file?

Vampire:
I pushed a version with that change, let's see whether something else breaks or you say this actually is necessary. :-)

Vampire:
Well, the tests passed with that change :-D

ben-manes:
It looks good to me. Running it against Caffeine and I don't see any of the hidden dependencies in the report (bundled ant, jacoco) and the undeclared category from [v39](https://github.com/ben-manes/gradle-versions-plugin/releases/tag/v0.39.0) has unit tests.

PR #736, opened 2023-02-27
Avoid configuring dependencyUpdates task

dreis2211:
Hi,

I noticed that the `dependencyUpdates` task (e.g. when added via global init script) is configured regardless of any task avoidance being used. This is caused by the usage of `tasks.findByName()` that was introduced with https://github.com/ben-manes/gradle-versions-plugin/pull/504 .
image
Eventually this causes an overhead by using Kotlin reflection in `DependencyUpdatesTask.callIncompatibleWithConfigurationCache`.

image

A workaround for this is to use `tasks.getNames().contains()`.

Let me know what you think.
Cheers,
Christoph


ben-manes:
thanks!

PR #747, opened 2023-03-14
Add github action for updating gradle wrapper

jaredsburrows:
@ben-manes 
Example: https://github.com/jaredsburrows/gradle-license-plugin/pull/257

### DIFF
--- a/.github/workflows/update-gradle-wrapper.yml
+++ b/.github/workflows/update-gradle-wrapper.yml
@@ -1,0 +1,22 @@
+name: Update Gradle Wrapper
+
+on:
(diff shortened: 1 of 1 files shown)


ben-manes:
This seems like it would get noisy? The gradle wrapper doesn't need to be updated very often. If it auto-merged supported automatically merging then that would be nice.

ben-manes:
It also bumped up the major version of Gradle, which may be incompatible. It's a nice idea but feels a bit burdensome?

jaredsburrows:
Yeah I'll close this.

Issue #748, opened 2023-03-14
Sonatype links are now broken because of search.maven.org


marcosalis:
First of all, thank you for this extremely useful plugin!

There is a slight issue with the reporting, as the Sonatype/Maven links to builds have changed.

Example of broken link in report:
`https://search.maven.org/artifact/androidx.activity/activity-ktx/1.6.1/bundle`

I believe that now `search.maven.org` has fully migrated to `central.sonatype.com`, as the message at [this link](https://central.sonatype.com/?smo=true) says.

ben-manes:
What is the reporting issue? That domain is only referenced as a (possible) quick link in the [html report](https://github.com/ben-manes/gradle-versions-plugin/blob/116a65ac9c42d7ee64789b7e3db30ba59c987e82/gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/reporter/HtmlReporter.kt#L419-L427). That android dependency is not on Maven Central and on Google's repository. That causes the link to fail as not found.

We don't know where the dependency was located by Gradle during resolution, it's just a best effort guess. The majority of the community uses Central, but some ecosystems decided to diverge. Many therefore use mvnrepository.com as a meta ui that indexes multiple repositories, but it is not official and displays ads which could erode trust.

marcosalis:
Yes indeed, Google's Android dependencies are on their repository ([here](https://maven.google.com/web/index.html?q=activity-ktx#androidx.activity:activity-ktx:1.6.1)).
I was about to suggest to use `mvnrepository.com` (the above dependency is [present](https://mvnrepository.com/artifact/androidx.activity/activity-ktx/1.6.1) there), but you are right in that it's not official.
It could be useful as a second link in the html report, but I understand if you prefer not adding it for the mentioned reasons.

> I believe that now search.maven.org has fully migrated to central.sonatype.com, as the message at [this link](https://central.sonatype.com/?smo=true) says.

This point is still relevant though, as at some point the `central.sonatype.com` might replace the currently used `search.maven.org` fully.

ben-manes:
I'd be fine having the urls swapped and an additional links to other repositories. It used to include jcenter in the html report. Having a few likely ones, including `mvnrepository`, is fine with me. I am hesitant to make it the only link since it is unofficial, but multiple quick links seems fine to me. Of course if most people would be happier with `mvnrepository` then I'm not strongly opposed.

[...]

PR #750, opened 2023-03-15
Fix broken Sonatype links in HTML reports

marcosalis:
https://github.com/ben-manes/gradle-versions-plugin/issues/748

Old `search.maven.org` domain was replaced with `central.sonatype.com`, the change was reflected in HTML generation code in `HtmlReporter.kt`.

Previous example link wasn't working anymore:
https://search.maven.org/artifact/com.azure/azure-core-http-netty/1.5.4

This URL leads to the correct page instead:
https://central.sonatype.com/artifact/com.azure/azure-core-http-netty/1.5.4


ben-manes:
Thanks!

PR #757, opened 2023-03-29
Update build.yml

jaredsburrows:
@ben-manes 

- `actions/checkout@v3` -> `actions/checkout@v3.5.0`
- `gradle/gradle-build-action@v2` -> `gradle/gradle-build-action@v2.4.0`
- Instead of always publishing via `always()`, let's use a real condition `github.repository == 'ben-manes/gradle-versions-plugin' && github.ref == 'refs/heads/master'` and remove the warning. Let it fail if there are no artifacts. Also, only publish on the master branch.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,7 +23,7 @@ jobs:
 
     steps:
       - name: Checkout Project
-        uses: actions/checkout@v3
+        uses: actions/checkout@v3.5.0
 
       - name: Validate Gradle Wrapper
         uses: gradle/wrapper-validation-action@v1.0.6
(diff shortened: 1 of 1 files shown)


ben-manes:
v2 / v3 will take the latest of that major version, so this should have no effect. It only pins up to that version, which will create dependabot noise. If pinning for reliability and security, then by hash is ideal even though more annoying. I don't think pinning is needed because these are core actions which will not break, so I think you can drop those changes.

ben-manes:
This isn't uploading a release artifact, only test results, so I think you can keep it. I don't see a harm if running on a fork and uploading the outputs for them to debug. They are small so minimal harm and no cache TTL is needed.

...

I guess that means I don't think any of these changes are helpful. Good ideas, but on inspection not very useful? I might be missing something and being dumb, though!

PR #768, opened 2023-04-17
Fix inherited collection to not be always empty

tresat:
This fixes a mistake in PR #721 which used `allDependencies` here instead of `dependencies`, causing the inherited set always be empty, as a result of subtracting `allDependencies` from it.

This also prevents inherited kotlin deps from being included in the latest set, to avoid issues with having the original and modified dependencies both added to the copy.


tresat:
FYI, reviewed this with @jvandort, who worked on the referenced PR and he agrees with the change.  

We also think it's worth adding some integration tests to ensure this continues to work, since this was a change in functionality that wasn't caught by existing tests, but I haven't attempted that here.

ben-manes:
Thanks a bunch!

PR #779, opened 2023-05-11
Replace deprecated toUpperCase() kotlin.text calls

smelfungus:
Hello! Kotlin's String `toUpperCase()` is deprecated. The new recommended way is to use `uppercase()`.


ben-manes:
thx!

PR #785, opened 2023-06-09
Allow filtering configurations

### LINKED ISSUE #782: Enhancement request to exclude configurations from check
Certain plugins create their own configurations for resolution, and these configurations are picked up by the plugin and merged into the report.  It causes some level of confusion in many circumstances when reading the reports.

For example, spotbugs plugin generates several configurations (Spotbugs, SpotbugsPlugins, SpotbugsSlf4j), which cause weirdness like duplicate entries for slf4j, entries for unknown libraries that are not directly configured, etc.  Similar things happen with com.diffplug.spotless, and other similar plugins.  

It would be good if it was possible to either specify the specific configurations to resolve, or specify configurations to exclude when executing the plugin.

bh-tt:
Closes #782

This needs someone with actual knowledge of kotlin/groovy to look at it, but the logic itself and the test is written.

### DIFF
--- a/README.md
+++ b/README.md
@@ -307,6 +307,19 @@ tasks.named<DependencyUpdatesTask>("dependencyUpdates").configure {
 
 Note: Do use the `plugins { .. }` syntax if you use the Kotlin DSL.
 
+#### Configuration filter
+You can change which dependency configurations the plugin checks for updates like this:
+
+```groovy
+// https://github.com/ben-manes/gradle-versions-plugin
+tasks.named("dependencyUpdates").configure {
+  filterConfigurations {
+    it.name.equals("runtimeClasspath") || it.name.equals("compileClasspath")
+  }
+}
+```
+
+
 #### Try out the samples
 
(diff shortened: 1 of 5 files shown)


ben-manes (approved):
Please fix the lint errors and then you can merge if the CI passes.

You are welcome to release by incrementing the version and creating release notes. That will create a tag and deploy the plugin.

Thanks!

bh-tt:
I should probably add a few lines to the README detailing the new feature as well?

ben-manes:
Oh yeah, that would be nice 🙂

bh-tt:
I dont mind being given release access to this repository, but could you be a little bit more careful giving someone who spent about an hour on a minor feature full write access to the repository? If I had malicious intentions it would be easy to add some malicious code to the plugin and release a new version. Your plugin is used in quite a few projects I believe which makes it a prime target for a supply chain attack.

I do not need write access to the main repository to create a PR so I should not have it.

ben-manes:
I appreciate the feedback. I think someone who discusses improvements on the issue tracker, contributes a feature, and writes tests should be given the benefit of being trusted. Otherwise I wouldn't have invited you as a contributor.

bh-tt:
Forgot to update the version, I've deleted the v0.47.0 tag and release and recreated it.

Might be good to add a 'how to release' section to the readme?

bwmeier:
Thank you for working on this @bh-tt !  I really appreciate the feature!

PR #807, opened 2023-09-06
Support gradle versions api url customization

sebek64:


ben-manes:
thanks, that seems more reasonable. You'd need to fix the broken tests, maybe add one if possible, and include a one-liner in the readme about the config.

sebek64:
I've fixed the failing tests and tuned README. I was thinking about how to write a new test - how to mock the gradle service on an alternative url. There are more ways. One is to start an embedded http server in the test. An alternative can be to put some file to this repo, and refer to this file on github.

It is up to you to chose the right way or suggest another one. Or if we can survive without a test.

ben-manes:
it looks like a large refactoring would be required to mock it with a file, even though that is quite convenient (like we do with the maven repository). Perhaps less so to spin up an http proxy, but that might feel like cumbersome busywork or be flaky.  This seems like a small enough change that it's not super important to test more than what we already have, so I'm fine merging as is. We can release it when the tests pass or whenever you're ready to use your enhancement.

ben-manes:
[releasing](https://github.com/ben-manes/gradle-versions-plugin/actions/runs/6106596552/job/16571964115), should be in the plugin repository in a few minutes.

sebek64:
Thank you, looking forward.

ben-manes:
and it's [live](https://plugins.gradle.org/plugin/com.github.ben-manes.versions). Thanks for the contribution!

Issue #424, opened 2020-09-08
Handle non-jar / metadata-only artifacts


sschuberth:
In regards to `com.github.ajalt.clikt:clikt`, this is because we don't handle non-jar dependencies yet. This dependency is [metadata only](https://repo1.maven.org/maven2/com/github/ajalt/clikt/clikt/3.0.1/), so an `extension` is applied during resolution. When we create our dependency query, we are not applying the classifier and extension. This causes a failure because Gradle is looking for a jar and fails.

I did find where to extract this information in #224 but never got around to adding it into the query. It's probably very simple if you'd like to give it a shot. It should just take a little formatting magic in the [query string](https://github.com/ben-manes/gradle-versions-plugin/blob/9effd941ff765dafb3149ac9246da5c7f288e333/src/main/groovy/com/github/benmanes/gradle/versions/updates/Resolver.groovy#L143).

_Originally posted by @ben-manes in https://github.com/ben-manes/gradle-versions-plugin/issues/334#issuecomment-688649827_

serebit:
0.31.0 still fails to detect *one* dependency in one of my projects, and that's `com.soywiz.korlibs.klock:klock`. Resolving this issue should fix that.

ben-manes:
@sschuberth your issue is due to JCenter which behaves oddly because it only half proxies queries. If you add Maven Central then it works fine. While it is only a metadata dependency, the declaration doesn't have an extension or classifier in the dependency declaration which is what we don't support (e.g. `a:b:1.0@zip`) yet.

@serebit Your issue appears to be due to needing the dependency capabilities copied over. That dependency has a platform capability (e.g. JVM) to select from. I would need a sample project to test against. The [api](https://docs.gradle.org/current/javadoc/org/gradle/api/artifacts/ModuleDependency.html) implies that we could provide the source's `getRequestedCapabilities()` via the target's `capabilities​(configureAction)`. Please provide a minimal failing test build that I can verify a fix against, and open a new issue with those details.

sschuberth:
> @sschuberth your issue is due to JCenter which behaves oddly because it only half proxies queries.

Would have have some more details on that, so we could report this to JCenter / JFrog?

ben-manes:
The error (`--info`) says that the metadata cannot be fetched, which a cURL verifies.
```
 - com.github.ajalt.clikt:clikt
[...]

PR #825, opened 2024-01-22
Ensure Kotlin dependencies are properly handled when present in a configuration with no parents

tresat:
This will handle the case of Kotlin plugins (applied to the parentless classpath configuration).

### DIFF
--- a/gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/Resolver.kt
+++ b/gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/Resolver.kt
@@ -89,11 +89,12 @@ class Resolver(
     revision: String,
     currentCoordinates: Map<Coordinate.Key, Coordinate>,
   ): Configuration {
-    // Kotlin deps anywhere in the hierarchy are a special case we'll handle later
-    val kotlinDeps = { dependency: ExternalDependency -> dependency.group == "org.jetbrains.kotlin" && dependency.version != null }
+    // Kotlin deps anywhere in the hierarchy are a special case we'll handle later separately, unless they are being
+    // forced, as is the case with plugins, in which case handle those deps here
+    val kotlinDeps = { dependency: ExternalDependency -> (dependency.group?.startsWith("org.jetbrains.kotlin") ?: false) && dependency.version != null }
     val latest = configuration.allDependencies
       .filterIsInstance<ExternalDependency>()
-      .filterNot(kotlinDeps)
+      .filterNot { kotlinDeps(it) && it.isForce }
       .mapTo(mutableListOf()) { dependency ->
         createQueryDependency(dependency as ModuleDependency)
       }
--- a/gradle-versions-plugin/src/test/groovy/com/github/benmanes/gradle/versions/PluginUpdateDetectionSpec.groovy
+++ b/gradle-versions-plugin/src/test/groovy/com/github/benmanes/gradle/versions/PluginUpdateDetectionSpec.groovy
@@ -1,0 +1,100 @@
+package com.github.benmanes.gradle.versions
+
+import org.gradle.testkit.runner.GradleRunner
+import org.junit.Rule
+import org.junit.rules.TemporaryFolder
+import spock.lang.See
+import spock.lang.Specification
+
+import static org.gradle.testkit.runner.TaskOutcome.SUCCESS
+
+final class PluginUpdateDetectionSpec extends Specification {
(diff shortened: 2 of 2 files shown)


ben-manes:
odd, the test suite breaks with this change

tresat:
You're right, this breaks additional tests.

I was getting failures in `DifferentGradleVersionsSpec` when I checked out `master` and ran `build` for the first time, so I assumed they weren't related.  Then I didn't run the full test suite after that, since it would always fail.

I'll take another look.

ben-manes:
iirc, that happens when you run with a newer jdk. You can use jenv, asdf, etc. to set a local one for the project. Trying right now with jdk11 on master to see if it passes locally. I don't recall what the issue was, but remember being confused by that test failure too

ben-manes:
yep, jdk11 worked locally

gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/Resolver.kt
```diff
@@ -89,11 +89,12 @@ class Resolver(
...
   ): Configuration {
-    // Kotlin deps anywhere in the hierarchy are a special case we'll handle later
-    val kotlinDeps = { dependency: ExternalDependency -> dependency.group == "org.jetbrains.kotlin" && dependency.version != null }
+    // Kotlin deps anywhere in the hierarchy are a special case we'll handle later separately, unless this configuration
+    // has no parents (such as the classpath conf, which contains plugins, including kotlin ones), in which case handle those deps here
+    val kotlinDeps = { dependency: ExternalDependency -> (dependency.group?.startsWith("org.jetbrains.kotlin") ?: false) && dependency.version != null }
     val latest = configuration.allDependencies
       .filterIsInstance<ExternalDependency>()
-      .filterNot(kotlinDeps)
+      .filter { configuration.hierarchy.size > 1 && kotlinDeps(it) }
```
ben-manes, line 97: maybe `filterNot`? trying locally

ben-manes: passes except for your new test, which I think might be a formatting difference?

```console
PluginUpdateDetectionSpec > kotlin plugin in the implementation configuration is properly handled FAILED
    Condition not satisfied:

    result.output.contains """The following dependencies have later milestone versions: - org.jetbrains.kotlin:kotlin-gradle-plugin [$KOTLIN_VERSION -> """
    |      |      |                                                                                                                   |
    |      |      false                                                                                                               1.6.0
    |
    |      > Task :dependencyUpdates
    |
    |      ------------------------------------------------------------
    |      : Project Dependency Updates (report to plain text file)
    |      ------------------------------------------------------------
    |
    |      The following dependencies are using the latest milestone version:
    |       - org.jetbrains.kotlin:kotlin-gradle-plugin:1.6.0
    |       - org.jetbrains.kotlin:kotlin-stdlib:1.6.0
```

tresat: I think detecting by `isForced` might be what's needed.

tresat:
A little more context: I saw the comments about plugins trying to force versions, so that led me to this path.  My approach to fixing this has been to preserve the special case logic for `inheritedKotlin` by making the smallest change that fixes these new test cases and leaving that alone.  

But after talking with @jvandort, it seems `isForce` isn't doing anything in newer versions of Gradle, and there is no way remaining for external deps to be marked `forced`.  So the effect of this change is to just always allow Kotlin deps to be included in that first loop that collects them into `latest`.

So I'm wondering if _any_ of the special casing for Kotlin is still necessary?  If I remove it all, I get the same passing build.  Perhaps it is no longer useful?  I couldn't find any existing tests that hinted at why it was still there.  Do you know how we could create one to confirm if it is needed?

ben-manes:
Here is the commit that introduced it, https://github.com/ben-manes/gradle-versions-plugin/commit/0a290fb86d8ba3962fead9d3c2be8fa91f1ab5e5, which seems to be trying to fix #424.

If we don't need it anymore that is better. wdyt?

ben-manes:
I released to unblock, but happy to revisit the removal of that code if its useless nowadays.

tresat:
@ben-manes Can you take a look at https://github.com/ben-manes/gradle-versions-plugin/pull/826 when you get a chance?  I think this is all that is needed.  I added tests for the kotlin stdlib dependency that was the origin of the special case in the `Resolver`.

PR #832, opened 2024-01-31
Add dependency-submission workflow and address deprecations/vulnerabilities

bigdaz:
- Updates workflows to use Node20-compatible actions where available
- Switches from `gradle/gradle-build-action` to `gradle/actions/setup-gradle`
- Adds a workflow using `gradle/actions/dependency-submission` to generate and upload dependency-graph
- Bumps the version of the 'ktlint' plugin to address reported vulnerability

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,10 +34,11 @@ jobs:
           distribution: temurin
           java-version: ${{ matrix.java-version }}
 
+      - name: Setup Gradle
+        uses: gradle/actions/setup-gradle@v3
+
       - name: Run build with caching enabled
-        uses: gradle/gradle-build-action@v3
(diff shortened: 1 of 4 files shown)


ben-manes:
oh, looks like a bunch of klint style issues. Is there a way to let it auto-fix?

ben-manes:
thanks as always!

bigdaz:
Sorry, I know absolutely zero about the `ktlint` plugin. I guess the new version has more checks?

I can remove that "fix" for now, so we get the rest of the changes through.

ben-manes:
oh no worries, I don't know it either. I found the its auto format task, fixed where it wouldn't do it automatically, and merged this all in. thanks again!

Issue #733, opened 2023-02-23
Cannot exclude kotlin-stdlib-jdk8 in 0.46.0


emilv:
After upgrade to the latest version 0.46.0 we get this error:

```
The following dependencies have later milestone versions:
 - org.jetbrains.kotlin:kotlin-stdlib-jdk8 [1.8.10 -> 1.8.20-Beta]
     https://kotlinlang.org/
```

Here is our normal way of excluding versions, inspired by your README:

```kotlin
        fun isNonStable(candidate: ModuleComponentIdentifier): Boolean {
            val version = candidate.version
            val stableKeyword = listOf("FINAL", "GA", "RELEASE").any { version.toUpperCase().contains(it) }
            val regex = "^[0-9,.v-]+(-r)?$".toRegex()
            val isStable = stableKeyword || regex.matches(version)
            return isStable.not().also { notStable ->
                if (notStable) {
                    logger.log(
                        LogLevel.WARN,
                        "Excluding ${candidate.module}:${candidate.version} from dependency check because: Not stable",
                    )
                }
            }
        }
        
        rejectVersionIf {
            isNonStable(candidate.version)
        }
```

The `isNonStable` check works in 0.45.0 (without the explicit exclude in excludeKotlinVersions), but not in 0.46.0.

It does trigger the warning in both 0.45.0 and 0.46.0:

```
Excluding kotlin-stdlib-jdk8:1.8.20-Beta from dependency check because: Not stable
Excluding kotlin-stdlib-jdk8:1.8.20-Beta from dependency check because: Not stable
```

But I still get the version warning:

```
The following dependencies have later milestone versions:
 - org.jetbrains.kotlin:kotlin-stdlib-jdk8 [1.8.10 -> 1.8.20-Beta]
     https://kotlinlang.org/
```

ben-manes:
I tried it on Groovy and Kotlin, and it worked for me. Can you make a minimal reproduction?

<details>
<summary>Kotlin</summary>

```kotlin
import com.github.benmanes.gradle.versions.updates.DependencyUpdatesTask

plugins {
    `java-library`
    id("com.github.ben-manes.versions") version "0.46.0"
}

repositories {
  mavenCentral()
}

fun String.isNonStable(): Boolean {
  val stableKeyword = listOf("RELEASE", "FINAL", "GA").any { toUpperCase().contains(it) }
  val regex = "^[0-9,.v-]+(-r)?$".toRegex()
[...]

PR #834, opened 2024-02-08
Documentation / Skip non-stable versions

johnjohndoe:
I tried to extend the rejection rule to exclude versions like `org.jetbrains.kotlin:kotlin-stdlib [1.9.22 -> 2.0.0-Beta3]` but for some reason it is still listed. 
Do you spot the mistake?

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,10 @@ no agreed standard on this, but this is a good starting point:
 
 ```groovy
 def isNonStable = { String version ->
+  def nonStableKeyword = ["BETA"].any { it -> version.toUpperCase().contains(it) }
+  if (nonStableKeyword) {
+    return true
+  }
   def stableKeyword = ['RELEASE', 'FINAL', 'GA'].any { it -> version.toUpperCase().contains(it) }
   def regex = /^[0-9,.v-]+(-r)?$/
   return !stableKeyword && !(version ==~ regex)


ben-manes:
You might see if my [build snippet](https://github.com/ben-manes/caffeine/blob/b100cb9bf01269fbd46e5eac7fabf5db2ceedd58/gradle/plugins/src/main/kotlin/lifecycle/dependency-versions-caffeine-conventions.gradle.kts) from Caffeine helps? I added an `unstableKeyword` and return that in the condition. I'm not sure why yours didn't work as the two look very similar in concept.

ben-manes:
trying your version in a groovy repl and it seems like it should work....

```groovy
def isNonStable = { String version ->
  def nonStableKeyword = ["BETA"].any { it -> version.toUpperCase().contains(it) }
  if (nonStableKeyword) {
    return true
  }
  def stableKeyword = ['RELEASE', 'FINAL', 'GA'].any { it -> version.toUpperCase().contains(it) }
  def regex = /^[0-9,.v-]+(-r)?$/
  return !stableKeyword && !(version ==~ regex)
}

println(isNonStable("2.0.0-Beta3"))
println(isNonStable("1.9.22"))
```

```console
true
false
```

johnjohndoe:
Yes, it seem to be correct. When I apply it to [my project](https://github.com/EventFahrplan/EventFahrplan/blob/master/gradle/versions.gradle) still the beta version is printed out when run the plugin task:

``` bash
 - org.jetbrains.kotlin:kotlin-stdlib [1.9.22 -> 2.0.0-Beta3]
     https://kotlinlang.org/
```

ben-manes:
I took that snippet into a sample project and it came out as you desired. This PR actually isn't needed because the keyword search is for stable or numeric, and anything else is assumed unstable to be excluded. So `beta` should have been dropped. However, I also see in your build it shows up which is weird, but I haven't looked into it much yet.

<details><summary>sample</summary>

```groovy
plugins {
  id 'com.github.ben-manes.versions' version '0.51.0'
  id 'java'
}

repositories {
  mavenCentral()
}

dependencyUpdates {
  def isNonStable = { String version ->
    def stableKeyword = ['RELEASE', 'FINAL', 'GA'].any {
      qualifier -> version.toUpperCase().contains(qualifier)
    }
    def regex = /^[0-9,.v-]+(-r)?$/
    return !stableKeyword && !(version ==~ regex)
  }
  rejectVersionIf {
    return isNonStable(it.candidate.version)
  }
}

dependencies {
  implementation 'org.jetbrains.kotlin:kotlin-stdlib:1.9.22'
}
```

```console
gradle dU --refresh-dependencies -q
executing gradlew instead of gradle

 - com.github.ben-manes.versions:com.github.ben-manes.versions.gradle.plugin:0.51.0
 - org.jetbrains.kotlin:kotlin-stdlib:1.9.22

 - Gradle: [8.0.2 -> 8.6]
```
</details>

ben-manes:
A println shows it rejections,
> Selection of org.jetbrains.kotlin:kotlin-stdlib:2.0.0-Beta3 rejected by component selection rule: Rejected by rejectVersionIf

but then I also observe,
> Comparing dependency (current: org.jetbrains.kotlin:kotlin-stdlib:2.0.0-Beta3, latest: 2.0.0-Beta3)

Which indicates a dynamic version somewhere on the declared dependencies, and that is somehow not applying the resolution strategy.

Unfortunately I'm not sure beyond that. It is a configuration in the `app` subproject, but it does not print out which one. I don't know why it is not being constrained by Gradle.

ben-manes:
oh, this is the regression in https://github.com/ben-manes/gradle-versions-plugin/issues/733. We might just want to revert the contributed changes (https://github.com/ben-manes/gradle-versions-plugin/issues/733#issuecomment-1451095661). I never had a chance to look into it and see if the contributor's improvements could be reworked to not cause this breakage. If I use v0.45 in your project then you get the right output.

I'll close pr. There is a workaround in that issue. It would be good to fix but it's not something I have the time to investigate unfortunately.

PR #856, opened 2024-03-29
Avoid spec-accepting overload of getFirstLevelModuleDependencies

jvandort:
This method will be deprecated soon in Gradle for removal in 9.0

The no-argument method will remain at least for the short term


ben-manes:
Thanks! I recall using the argument for compatibility. Amusing to find that old [commit](https://github.com/ben-manes/gradle-versions-plugin/commit/076361d83099fa452e83f0c3b9604974d3361a35) 

> **Remove guard clause due to bad method invocation.**
The no-arg `getFirstLevelModuleDependencies()` method is only availabe
on the default implementation and is not defined on the interface for
LenientConfiguration. The anonymous implementation returned from when
any empty dependency set is used lacks this helper method, resulting
in a runtime failure. The interface method is now invoked with a
filter to collect everything.

PR #879, opened 2024-07-30
Fix various warnings

jaredsburrows:
task -> task.register
all -> configureEach

### DIFF
--- a/gradle-versions-plugin/build.gradle
+++ b/gradle-versions-plugin/build.gradle
@@ -2,7 +2,7 @@ plugins {
   id 'org.jetbrains.kotlin.jvm'
   id 'org.jetbrains.dokka'
   id 'org.jlleitschuh.gradle.ktlint'
-  id 'com.gradle.plugin-publish' version '0.21.0'
+  id 'com.gradle.plugin-publish'
   id 'maven-publish'
   id 'java-gradle-plugin'
   id 'java-library'
@@ -13,7 +13,7 @@ group = GROUP
 version = VERSION_NAME
 
 // Write the plugin's classpath to a file to share with the tests
-task createClasspathManifest {
+tasks.register('createClasspathManifest') {
   def outputDir = file("$buildDir/$name")
 
   inputs.files sourceSets.main.runtimeClasspath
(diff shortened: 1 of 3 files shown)


gradle-versions-plugin/build.gradle
```diff
@@ -41,14 +50,16 @@ dependencies {
...
+  dependsOn dokkaHtml
   group = 'Publications'
   description = 'Create jar of documentation.'
   archiveClassifier = 'javadoc'
   from dokkaJavadoc.outputDirectory
 }
 
-task sourcesJar(type: Jar, dependsOn: classes) {
+tasks.register('sourcesJar', Jar) {
+  dependsOn classes
```
ben-manes, line 62: I think nowadays one can use `java.withSourcesJar()`

gradle-versions-plugin/build.gradle
```diff
@@ -41,14 +50,16 @@ dependencies {
   testImplementation 'org.spockframework:spock-junit4:2.1-M2-groovy-3.0', { exclude module: 'groovy-all' } // Use localGroovy()
 }
 
-task docsJar(type: Jar, dependsOn: dokkaHtml) {
+tasks.register('docsJar', Jar) {
+  dependsOn dokkaHtml
```
ben-manes, line 54: and here I think can be done via `java.withJavadocJar()`

jaredsburrows: I think we have this custom task for dokka.

gradle-versions-plugin/build.gradle
jaredsburrows: I can move us to use https://github.com/vanniktech/gradle-maven-publish-plugin to help remove these custom tasks in the future.

ben-manes (approved):

gradle-versions-plugin/build.gradle
jaredsburrows: @ben-manes Actually, we are only using Gradle's publish plugin - https://docs.gradle.org/current/userguide/publishing_gradle_plugins.html we are not publishing the docs or sources. Looks like we have not published to maven in a long time? https://mvnrepository.com/artifact/com.github.ben-manes/gradle-versions-plugin

ben-manes: Oh right, it predated the plugin portal by a few years.

PR #880, opened 2024-07-30
gradle 8.9

jaredsburrows:


ben-manes (approved):

jaredsburrows:
@ben-manes 

Might have to hold off on this - `Caused by: java.lang.IllegalArgumentException: Unsupported class file major version 65`

ben-manes:
I suppose using toolchains for the daemon and language version might help? Then 21 could be used by Gradle itself while tests run on Java 8. We could use `--release` or use the jdk for the correct byte code version. If it ever gets to be a pain upgrading is probably fine.

jaredsburrows:
Need to focus on https://github.com/ben-manes/gradle-versions-plugin/pull/895

PR #891, opened 2024-07-30
Fix various warnings

jaredsburrows:
Fix various warnings such as:

```
w: file:///Users/<>/repo/gradle-versions-plugin/gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/DependencyUpdatesTask.kt:127:66 Java type mismatch: inferred type is 'groovy.lang.Closure<*>', but 'groovy.lang.Closure<kotlin.Any!>' was expected.
```

### DIFF
--- a/gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/VersionsPlugin.kt
+++ b/gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/VersionsPlugin.kt
@@ -17,7 +17,7 @@ class VersionsPlugin : Plugin<Project> {
     requireSupportedSaxParser()
 
     val tasks = project.tasks
-    if (!tasks.getNames().contains("dependencyUpdates")) {
+    if (!tasks.names.contains("dependencyUpdates")) {
       tasks.register("dependencyUpdates", DependencyUpdatesTask::class.java)
     }
   }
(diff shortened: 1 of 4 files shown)


gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/DependencyUpdates.kt
```diff
@@ -21,18 +21,18 @@ import org.gradle.api.specs.Spec
 class DependencyUpdates
   @JvmOverloads
   constructor(
-    val project: Project,
-    val resolutionStrategy: Action<in ResolutionStrategyWithCurrent>?,
-    val revision: String,
+    private val project: Project,
```
ben-manes, line 24: I’m doubtful, but an extension plugin could have used this constructor if public. It might not be worth worrying about though

ben-manes (approved):

gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/DependencyUpdates.kt
ben-manes: https://github.com/fkorotkov/gradle-libraries-plugin/blob/92311acf0cdd7019f7ba7fb55d7c70e98acde9ec/src/main/java/com/fkorotkov/gradle/libraries/tasks/LibrariesBaseTask.java#L47

jaredsburrows: Reverted.

PR #916, opened 2024-11-26
Use JUnit platform

jaredsburrows:
Use JUnit platform

@ben-manes


jaredsburrows:
@ben-manes Tests are good 👍

PR #918, opened 2024-11-26
Show test report

jaredsburrows:
Make sure tests are counted and actually ran

See https://github.com/ben-manes/gradle-versions-plugin/pull/916

@ben-manes


jaredsburrows:
@ben-manes This should fail the first time. Once https://github.com/ben-manes/gradle-versions-plugin/pull/916 is merged, I will rebase and rerun this.

PR #936, opened 2025-02-21
Updated plugin with autoupdater and group exclusion for build file

rohan04:
Sample gradle command: 

./gradlew dependencyUpdates autoUpdateDependencies -PallowMajorUpdates=true -PallowMinorUpdates=true -PignoredGroups=com.google.guava,org.springframework

Argument                              | Description

-PallowMajorUpdates=true | Enables major version updates
-PallowMinorUpdates=true | Enables minor version updates
-PignoredGroups=com.google.guava,org.springframework | Specifies which dependency groups to ignore

### DIFF
--- a/gradle-versions-plugin/build.gradle.kts
+++ b/gradle-versions-plugin/build.gradle.kts
@@ -1,6 +1,7 @@
 plugins {
   alias(libs.plugins.kotlin.jvm)
-  alias(libs.plugins.dokka)
+  alias(libs.plugins.dokka) apply false // If using Version Catalogs
+  // id("org.jetbrains.dokka") version "1.9.20" apply false // Latest stable
   alias(libs.plugins.ktlint)
   alias(libs.plugins.plugin.publish)
   alias(libs.plugins.versions)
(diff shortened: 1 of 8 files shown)


rohan04:
@ben-manes I have added some extra functionality to the plugin to update the dependencies automatically within the gradle file and some other features mentioned above. pls review.

ben-manes:
Hi @rohan04!

Thanks for the offer but I think this would be best done in your own plugin. We'd be happy to list that with the other [related plugins](https://github.com/ben-manes/gradle-versions-plugin?tab=readme-ov-file#related-plugins).

There are many ways to define dependencies in Gradle. For some it is inline as your regex handles, whereas others will use variables to build a formatted string. That might be version constants in build file or external definitions, like the old `dependencies.gradle` manifest idiom. Some have used a custom file format to organize around (e.g. a .properties file). A variant of that are large organizations on Maven that would publish internal BOMs for all of their teams to consume, so they like to keep that approach when migrating to Gradle for newer projects. This means that every project has a unique approach in how they managed versions, so Gradle introduced version catalogs as their preferred one to coerce the community towards a standard style.

That is all quite messy so we have left it to others to sort out their preferred project management. The related plugins were built on top of ours to provide a rich, opinionated solution. You might want to review those to see if any would work for you, whether you'd like to contribute to them, or if you'd like to add your own to the options.

rohan04:
@ben-manes Thanks, I'll create a seperate plugin, appreciate the response.

PR #955, opened 2025-07-15
Update README.md for upstream Gradle issue workaround

elevenfive:
Add known issues section and workaround for configuration cache incompatibility.

```
5 problems were found storing the configuration cache, 4 of which seem unique.
- Task `:app:dependencyUpdates` of type `com.github.benmanes.gradle.versions.updates.DependencyUpdatesTask`: cannot deserialize Gradle script object references as these are not supported with the configuration cache.
  See https://docs.gradle.org/8.14.3/userguide/configuration_cache.html#config_cache:requirements:disallowed_types
- Task `:app:dependencyUpdates` of type `com.github.benmanes.gradle.versions.updates.DependencyUpdatesTask`: cannot serialize Gradle script object references as these are not supported with the configuration cache.
  See https://docs.gradle.org/8.14.3/userguide/configuration_cache.html#config_cache:requirements:disallowed_types
- Task `:app:dependencyUpdates` of type `com.github.benmanes.gradle.versions.updates.DependencyUpdatesTask`: cannot serialize a lambda that captures or accepts a parameter of type 'org.gradle.api.artifacts.Configuration' as these are not supported with the configuration cache.
  See https://docs.gradle.org/8.14.3/userguide/configuration_cache.html#config_cache:requirements:disallowed_types
- Task `:app:dependencyUpdates` of type `com.github.benmanes.gradle.versions.updates.DependencyUpdatesTask`: invocation of 'Task.project' at execution time is unsupported.
  See https://docs.gradle.org/8.14.3/userguide/configuration_cache.html#config_cache:requirements:use_project_during_execution
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@ checks for updates to Gradle itself.
   - [buildscript block](#buildscript-block)
   - [Using a Gradle init script](#using-a-gradle-init-script)
   - [Related plugins](#related-plugins)
+  - [Known issues](#known-issues)
 - [dependencyUpdates](#dependencyupdates)
   - [Multi-project build](#multi-project-build)
   - [Revisions](#revisions)
@@ -121,6 +122,9 @@ You may also wish to explore additional functionality provided by,
  - [gradle-update-notifier](https://github.com/y-yagi/gradle-update-notifier)
  - [refreshVersions](https://github.com/jmfayard/refreshVersions)
  - [update-versions-gradle-plugin](https://github.com/tomasbjerre/update-versions-gradle-plugin)
+
+### Known Issues ###
+[Gradle configuration cache](https://docs.gradle.org/current/userguide/configuration_cache.html) is not supported.  To work around the issue, run with `--no-configuration-cache` if it is enabled in your Gradle configuration.
 
 ## Tasks
 
--- a/README.md
+++ b/README.md
@@ -123,8 +123,8 @@ You may also wish to explore additional functionality provided by,
  - [refreshVersions](https://github.com/jmfayard/refreshVersions)
  - [update-versions-gradle-plugin](https://github.com/tomasbjerre/update-versions-gradle-plugin)
 
-### Known Issues ###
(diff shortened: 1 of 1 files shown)


elevenfive:
Unfortunately it's not just an INFO, it's a silent failure:

```
> Task :app:dependencyUpdates

:app Project Dependency Updates (report to plain text file)

No dependencies found.
```

ben-manes:
hmm, I don't have that because it should be automatically disabled in the current versions thanks to the [notCompatibleWithConfigurationCache](https://github.com/ben-manes/gradle-versions-plugin/blob/0e8aff41cfdc8297cc21a13bb4cf568e2520b8f7/gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/DependencyUpdatesTask.kt#L187-L190) declaration. 

When I run it on a [multi-project build](https://github.com/ben-manes/caffeine) it resolves properly. However, I run it at the root project and let it create an aggregated report of all child projects. If you are using Gradle 9.0-rc then their changes require a `--no-parallel` due to disallowing resolution of the project configurations. You can also look at `--info` to see if your configurations failed for that or a similar problem ([error logging](https://github.com/ben-manes/gradle-versions-plugin/blob/0e8aff41cfdc8297cc21a13bb4cf568e2520b8f7/gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/DependencyUpdates.kt#L89-L100)).

elevenfive:
Using Gradle `8.14.3` in a simple project with two modules, one dependent on the other.

Screenshot from `--info`:

Screenshot from 2025-07-15 16-13-02

ben-manes:
The problem report is still kind of fake because its only showing those as warnings. If the configuration cache was actually enabled then the build would fail, not just log an error. You want to see the errors in the info log which are not exposed to that report, for whatever reason. You should see an exception for the log message `Skipping configuration`

elevenfive:
Hm!  Running with `./gradlew dependencyUpdates --debug --stacktrace` also shows no exception(s), and no instance of `skipping configuration`.

ben-manes:
right, we already opt out inside the plugin. That is why I am so confused. Are you using the current release?

elevenfive:
Ah.  Yes, on latest release `0.52.0`

PR #984, opened 2026-03-17
Fix ConcurrentModificationException when iterating configurations

elevenfive:
## Summary
- Copies the configurations `Set` to a `List` before iterating in `DependencyUpdates.resolveProjects()` to prevent `ConcurrentModificationException` when configurations are modified during resolution.

## Test plan
- [ ] Verify existing tests pass with `./gradlew test`
- [ ] Confirm the fix prevents `ConcurrentModificationException` in projects where configurations are added during resolution


elevenfive:
@ben-manes I noticed CME after upgrading to Gradle 9.4 today.  This resolves the issue.

ben-manes:
Sorry, been meaning to get to this. There was some earlier discussions and seems related to Android misuse and Gradle changes to lazy configurations. I'll try to revisit it this week and get a patch. Else you can downgrade a few versions when we used to have this ordered differently, but switched due to other Gradle/plugins breaking it that seems to not happen on the latest Gradle.

ben-manes:
Released in v0.54.0. Thank you for the patience and support!

PR #927, opened 2025-01-22
Fix CME, update dependencies

elevenfive:
- See: https://github.com/ben-manes/gradle-versions-plugin/issues/907

PR #985, opened 2026-04-19
fix: use copied configurations to workaround issues with plugins adding dynamic configs

### LINKED ISSUE #930: dependencyUpdates task fails with ConcurrentModificationException since 0.52.0 in certain projects
I am getting a `ConcurrentModificationException` when running the `dependencyUpdates` task in some of my Spring Boot projects after upgrading them to `gradle-versions-plugin` v0.52.0.  After some experimentation, I've narrowed it down to projects which use all 3 of these plugins:

- `io.spring.dependency-management`, any version since at least v3.1.0
- `info.solidsoft.pitest` v1.15.0
- `com.github.ben-manes.versions` v0.52.0

The relevant bit of the stack trace (I can copy the whole thing here if you'd like):

```
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':dependencyUpdates'.
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.lambda$executeIfValid$1(ExecuteActionsTaskExecuter.java:130)
        at org.gradle.internal.Try$Failure.ifSuccessfulOrElse(Try.java:293)
        <snip>
Caused by: java.util.ConcurrentModificationException
        at org.gradle.api.internal.collections.FilteredElementSource$FilteringIterator.findNext(FilteredElementSource.java:115)
        at org.gradle.api.internal.collections.FilteredElementSource$FilteringIterator.next(FilteredElementSource.java:134)
        at org.gradle.api.internal.DefaultDomainObjectCollection$IteratorImpl.next(DefaultDomainObjectCollection.java:494)
        at com.github.benmanes.gradle.versions.updates.DependencyUpdates.resolveProjects(DependencyUpdates.kt:78)
        at com.github.benmanes.gradle.versions.updates.DependencyUpdates.run(DependencyUpdates.kt:46)
        at com.github.benmanes.gradle.versions.updates.DependencyUpdatesTask.dependencyUpdates(DependencyUpdatesTask.kt:140)
        at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
        at org.gradle.internal.reflect.JavaMethod.invoke(JavaMethod.java:125)
        at org.gradle.api.internal.project.taskfactory.StandardTaskAction.doExecute(StandardTaskAction.java:58)
        <snip>
```

Crashing line: [DependencyUpdates.kt:78](https://github.com/ben-manes/gradle-versions-plugin/blob/v0.52.0/gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/DependencyUpdates.kt#L78)

I have a minimal reproduction here: <https://github.com/mitchellmebane/gradle-versions-plugin-crash/>

This sounds like it might have been caused by the fix to #907, or perhaps that fix just moved the weak point into the loop in `resolveProjects` and changed the timing enough that my projects only hit it now.

I don't know much about Gradle internals, but I'll try and dig into this more if I have time.

chadlwilson:
- fixes #930

Reverts to the earlier approach prior to #927, but without calling `.size` which seemingly causes the configurations to be resolved and the same side effect of adding new configurations, and triggers the same problem as earlier.

I did not specifically test it with pitest and Spring Dependency Management; just worked on the assumption that if the same small change broke them, then reverting to how it was before (minus the pre-sizing) is likely sufficient for them also.

I did actually have a test that failed with the `ConcurrentModificationException` before and passed afterwards (On Gradle 8.4; later plugin versions need later Gradle versions), but this Quarkus version requires Java 17, and this plugin seems to run tests only against Java 8 + 11 and doesn't currently seem to have any infrastructure to test against various Java versions and ignore tests on certain Java versions - so I removed it.

```groovy
  @Issue("https://github.com/ben-manes/gradle-versions-plugin/issues/930")
  def "plugins that add configurations dynamically are properly handled"() {
    given:
    testProjectDir.newFile('build.gradle') <<
      """
        plugins {
            id 'com.github.ben-manes.versions'
            id 'io.quarkus' version '3.25.4'
        }
      """.stripIndent()

    when:
    def result = GradleRunner.create()
      .withProjectDir(testProjectDir.root)
      .withArguments('dependencyUpdates', '--info')
      .withPluginClasspath()
      .build()

    then:
    result.output.find(/The following dependencies have later milestone versions:\n - io\.quarkus:io\.quarkus\.gradle\.plugin \[3\.25\.4 -> /)
  }
```

### DIFF
--- a/gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/DependencyUpdates.kt
+++ b/gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/DependencyUpdates.kt
@@ -41,13 +41,13 @@ class DependencyUpdates
     fun run(): DependencyUpdatesReporter {
       val projectConfigs =
         project.allprojects
-          .associateBy({ it }, { it.configurations.matching(filterConfigurations) })
+          .associateBy({ it }, { it.configurations.matching(filterConfigurations).toCollection(LinkedHashSet()) })
 
       val status: Set<DependencyStatus> = resolveProjects(projectConfigs, checkConstraints)
 
       val buildscriptProjectConfigs =
         project.allprojects
-          .associateBy({ it }, { it.buildscript.configurations })
+          .associateBy({ it }, { it.buildscript.configurations.toCollection(LinkedHashSet()) })
       val buildscriptStatus: Set<DependencyStatus> =
         resolveProjects(
           buildscriptProjectConfigs,


chadlwilson:
Sorry, this is an alternative to https://github.com/ben-manes/gradle-versions-plugin/pull/984 which I did not see when creating this, since it was not linked to the issue. Not to worry. The other one will probably work - but this one uses a `LinkedHashSet` just in case some later code was expecting set semantics from the collection, as with `NamedDomainObjectCollection`s.

Feel free to throw away if it's not useful / not solid enough or you prefer the other approach.

ben-manes:
Thanks, I'll take this one since its a direct restore of the prior behavior

Issue #906, opened 2024-09-17
Plugin doesn't respect `rejectVersionIf` lambda


rynkowsg:
I made a custom filter, that does two things:
- does not update non-stable after stable
- ensure that guava versions change is rejected if they are of different flavors

Here is the convention plugin containing entire logic _(skip reading it and scroll down)_:
```kotlin
import GuavaCheckResult.GUAVA_CANDIDATE_NEWER
import GuavaCheckResult.GUAVA_CANDIDATE_OLDER
import GuavaCheckResult.GUAVA_DIFFERENT_FLAVORS
import GuavaCheckResult.GUAVA_SAME_VERSIONS
import GuavaCheckResult.NOT_GUAVA
import buildcfg.GradlePluginId
import com.github.benmanes.gradle.versions.updates.DependencyUpdatesTask
import com.github.benmanes.gradle.versions.updates.resolutionstrategy.ComponentSelectionWithCurrent
import org.gradle.api.Plugin
import org.gradle.api.Project
import org.gradle.kotlin.dsl.withType

val isDEBUG: Boolean = System.getenv("DEBUG").let {
    println("DEBUG=$it")
    it in listOf("true", "1") }

fun log(message: String) {
    if (isDEBUG) {
        println(message)
    }
}

private data class VersionComparisonData(
    val group: String?,
    val module: String?,
    val oldVersion: String,
    val newVersion: String,
)

class VersionsConventionPlugin : Plugin<Project> {
    override fun apply(project: Project) = with(project) {
        println("Applying VersionsConventionPlugin to $name")
        pluginManager.apply(GradlePluginId.VERSIONS)

        tasks.withType<DependencyUpdatesTask>().configureEach {
            checkForGradleUpdate = true
            outputFormatter = "plain,html"
            outputDir = "build/dependencyUpdates"
            reportfileName = "report"
            val filter: ComponentSelectionWithCurrent.() -> Boolean = {
                val candidate = this
                val comparisonData = VersionComparisonData(
                    group = candidate.candidate.group,
                    module = candidate.candidate.module,
                    oldVersion = candidate.currentVersion,
                    newVersion = candidate.candidate.version,
                )
                log("comparisonData: $comparisonData")
                val shouldReject = !shouldApprove(comparisonData)
                log("shouldReject: $shouldReject")
                log("------------------------------------------------------------")
                shouldReject
            }
            rejectVersionIf(filter)
        }
    }
}

[...]

Issue #348, opened 2019-10-01
New resolution rules that take the current version into account do not work properly for dependencies with multiple versions


Vampire:
I have Spotbugs with multiple version in different configurations.
The cause for this is, that I use the latest stable release 3.1.12 for the actual analyzing, but I use the overhauled stylesheet from 4.0.0-beta4 to generate the HTML report.

If I follow the "Example 1" style:
```kotlin
rejectVersionIf {
    val preliminaryReleaseRegex = Regex("""(?i)[.-](?:${listOf(
            "alpha",
            "beta",
            "..."
    ).joinToString("|")})[.\d-]*""")
    preliminaryReleaseRegex.containsMatchIn(candidate.version)
}
```
and am one minor version behind for both versions, I'm correctly advised to update the stable version but - also correctly as pre releases are rejected - not to update the beta version:
```
The following dependencies exceed the version found at the milestone revision level:
 - com.github.spotbugs:spotbugs [4.0.0-beta3 <- 3.1.12]
     https://spotbugs.github.io/

The following dependencies have later milestone versions:
 - com.github.spotbugs:spotbugs [3.1.11 -> 3.1.12]
     https://spotbugs.github.io/
```

I thought following the new "Example 2" style would be the solution:
```kotlin
rejectVersionIf {
    val preliminaryReleaseRegex = Regex("""(?i)[.-](?:${listOf(
            "alpha",
            "beta",
            "..."
    ).joinToString("|")})[.\d-]*""")
    preliminaryReleaseRegex.containsMatchIn(candidate.version)
            && !preliminaryReleaseRegex.containsMatchIn(currentVersion)
}
```
But unfortunately I'm not as exepcted advised to update the stable version to 3.1.12 and the unstable version to 4.0.0-beta4, but both to the beta.
```
The following dependencies have later milestone versions:
 - com.github.spotbugs:spotbugs [3.1.11 -> 4.0.0-beta4]
     https://spotbugs.github.io/
 - com.github.spotbugs:spotbugs [4.0.0-beta3 -> 4.0.0-beta4]
     https://spotbugs.github.io/
```

As the component selection now is calculated relative to the current version, it would be nice if it would also be applied relative to the current version and not for all configurations the same.

cc @ghus-raba

ben-manes:
[...]

PR #988, opened 2026-07-19
Run dependency actions before reading the declared set (fixes #987)

### LINKED ISSUE #987: Dependencies contributed by `withDependencies`/`defaultDependencies` are reported at a fabricated version and skip `rejectVersionIf`
---

A dependency that a plugin adds to a configuration lazily is reported as if the version it resolved to were the version the build declares. The report says it is using the latest milestone version when it is not, `rejectVersionIf` is never called for it (so a rule that rejects unstable candidates has no effect on that module), and in a build where the same module is declared normally somewhere else it can appear in two report groups at once, listed as up to date under one version and outdated under another. Because the outcome depends on whether something else in the build resolved the configuration first, the same build can report differently between runs.

This is how I first hit it: reports listing beta versions for `kotlin-stdlib`, contributed lazily by the Kotlin plugin, in a build whose `rejectVersionIf` rejects unstable candidates.

Related to #906, which shows the same report signature from a different cause.

## Reproducer

`settings.gradle`:

```groovy
rootProject.name = 'repro'
```

`build.gradle`:

```groovy
plugins {
  id 'java'
  id 'com.github.ben-manes.versions' version '0.54.0'
}

repositories { mavenCentral() }

// Stand-in for what the Kotlin plugin does for kotlin-stdlib, and the
// java/jacoco/checkstyle/pmd plugins do for their tool versions.
configurations.compileClasspath.withDependencies { deps ->
  deps.add(project.dependencies.create('com.google.guava:guava:30.0-jre'))
}

tasks.named('dependencyUpdates').configure {
  checkForGradleUpdate = false
  rejectVersionIf {
    println("rejectVersionIf: ${it.candidate.group}:${it.candidate.module} ${it.currentVersion} -> ${it.candidate.version}")
    false
  }
}
```

Expected: guava listed as outdated, `30.0-jre -> 33.6.0-jre`, and `rejectVersionIf` invoked for it.

Actual (plugin 0.54.0, Gradle 8.4):

```
rejectVersionIf: com.github.ben-manes.versions:com.github.ben-manes.versions.gradle.plugin 0.54.0 -> 0.54.0

 - com.github.ben-manes.versions:com.github.ben-manes.versions.gradle.plugin:0.54.0
 - com.google.guava:guava:30.0-jre
```

guava 30.0-jre is reported as the latest milestone version, and `rejectVersionIf` is never invoked for it: the only line printed is for the plugin marker artifact.

ianbrandt:
Fixes #987.

`Resolver.resolve` reads the declared set through `getCurrentCoordinates` before anything resolves, but `withDependencies` and `defaultDependencies` actions run at resolution time. A dependency contributed that way is missing from `declared`, gets dropped by `coordinates.keys.retainAll(declared.keys)`, and then reaches two places that tolerate a missing current coordinate: `wrapComponentSelection` returns null, so `rejectVersionIf` never runs for it, and `getStatus`'s `originalCoordinate ?: resolvedCoordinate` reports the resolved version as the declared one.

One statement at the top of `resolve`:

```kotlin
configuration.incoming.dependencies
```

`getDependencies()` runs the dependency actions across the `extendsFrom` hierarchy without triggering graph resolution.

### Report output changes

Coordinates the Kotlin plugin contributes lazily were previously compared against themselves, so they read as up to date. They now get a real latest-version query. With Kotlin 1.7.0 declared:

- `kotlin-compiler-embeddable` and `kotlin-klib-commonizer-embeddable` move from current to outdated. They were the only current entries in that scenario, so its "using the latest milestone version" section disappears.
- `kotlin-stdlib-jdk8` shows up as outdated; previously it was absent from the report.

Nobody updates these coordinates directly; they follow the KGP version. But KGP declares them on the project's configurations, they are genuinely stale, and `rejectVersionIf` now runs for them, so builds that consider them noise can filter them like any other dependency, which previously wasn't possible. That last part is what #987 grew out of: reports listing beta versions for `kotlin-stdlib` in a build whose `rejectVersionIf` rejects unstable candidates, because the rule was silently skipped for it. Tool versions from `defaultDependencies` (jacoco, checkstyle, pmd) change the same way: previously pinned at the default and never queried, now resolved like anything else.

Two `KotlinDependencyUpdatesSpec` expectations widen to match. `LazyDependencySpec` is new and covers both contribution routes. Five of the seven cases across the two specs are red without the `Resolver` change.

Let me know if the output changes are acceptable, or if you want it handled differently.

### DIFF
--- a/gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/Resolver.kt
+++ b/gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/Resolver.kt
@@ -53,6 +53,12 @@ class Resolver(
     configuration: Configuration,
     revision: String,
   ): Set<DependencyStatus> {
+    // Runs the actions that contribute dependencies lazily, so that the declared set below is read
+    // after them rather than before. A contribution missing from that set is discarded as an
+    // undeclared dependency, or reported with the resolved version as its declared one.
+    // https://github.com/ben-manes/gradle-versions-plugin/issues/987
+    configuration.incoming.dependencies
+
     val coordinates = getCurrentCoordinates(configuration)
     val latestConfiguration = createLatestConfiguration(configuration, revision, coordinates)
     val lenient = latestConfiguration.resolvedConfiguration.lenientConfiguration
--- a/gradle-versions-plugin/src/test/groovy/com/github/benmanes/gradle/versions/KotlinDependencyUpdatesSpec.groovy
+++ b/gradle-versions-plugin/src/test/groovy/com/github/benmanes/gradle/versions/KotlinDependencyUpdatesSpec.groovy
@@ -56,7 +56,14 @@ final class KotlinDependencyUpdatesSpec extends Specification {
       .build()
 
     then:
-    result.output.find(/The following dependencies have later milestone versions:\n - org\.jetbrains\.kotlin:kotlin-gradle-plugin \[$DECLARED_KOTLIN_VERSION -> 2\..*\]/)
+    def inheritedByJvmPlugin =
+      / - org\.jetbrains\.kotlin:kotlin-compiler-embeddable \[$DECLARED_KOTLIN_VERSION -> 2\..*\]\n\s+https:\/\/kotlinlang\.org\/\n/
+    result.output.find(/The following dependencies have later milestone versions:\n/ +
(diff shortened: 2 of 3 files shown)


ben-manes:
Thanks!

ianbrandt:
Thanks for the merge! I may have a common fix for #906 and #348 as well. I'll file a PR for those if so.

Issue #65, opened 2015-06-25
Confusing report if different versions of a library are used in root and subproject


jochenberger:
Extracting from #63.
I have a project with a subproject. The root project depends on version 0.1.2 of a library and the subproject depends on 0.3.0, which is the latest version. The versions plugin is applied to the root project only.
0.11.3 lists the version as upgradeable [0.3.0->0.3.0], 0.10.x listed it as up-to-date (0.3.0).

jochenberger:
The library is added to the VersionMapping twice.

ben-manes:
It should be shown twice (#62), but one listed as upgradable and the other as up-to-date. Obviously that's not always working, though I swore my test listed them correctly.

ben-manes:
Please test and reopen if this is still broken. I made the tests explicit instead of simply checking the counts.

ben-manes:
Reopening after visual inspecting the report, which is screwy.

ben-manes:
Okay, now looks right after tweaking the reporter. That area needs a lot of cleanup, which I'll do after things settle down in the resolution logic.

jochenberger:
I'm afraid it's not entirely fixed. Meanwhile, version 0.4.0 of the library is available. The root project is still at 0.1.2 and the subproject is at 0.3.0. Not, the plugin says `[0.3.0 -> 0.4.0]` for both the root and the subproject.

ben-manes:
Could you put a simple build file example that I can test with?

I'm tweaking the below script with different versions. If the dependency maps to different groups then I see it listed individually as desired. If they fit in the same group, e.g. upgrade 15, 16-rc-1 --> 18, then only one is listed. That indicates that the analysis / version mapping is correct, but there is probably a map in the reporter that needs to be replaced with a multi-map.

```gradle
apply plugin: 'com.github.ben-manes.versions'

buildscript {
  repositories {
    jcenter()
    mavenLocal()
  }

  dependencies {
    classpath 'com.github.ben-manes:gradle-versions-plugin:+'
    classpath 'com.google.guava:guava:15.0'
  }
}

configurations {
  upgradesFound
}

dependencies {
  upgradesFound 'com.google.guava:guava:16.0-rc1'
}
```

ben-manes:
I updated the example's [build.gradle](https://github.com/ben-manes/gradle-versions-plugin/blob/master/examples/build.gradle) to experiment with. I think my changes fix this.

jochenberger:
[...]

PR #989, opened 2026-07-19
Report each declared version against the latest found for it (fixes #348) (fixes #906)

### LINKED ISSUE #348: New resolution rules that take the current version into account do not work properly for dependencies with multiple versions

### LINKED ISSUE #906: Plugin doesn't respect `rejectVersionIf` lambda

ianbrandt:
Fixes #348 and fixes #906.

`VersionMapping.organize()` joins declared versions to latest versions with `latest.associateBy { it.key }`, keyed on group and name alone, so a module declared at two versions gets one latest for both. That mis-buckets the older declaration and prints another declaration's upgrade target on its row, though each `DependencyStatus` already carries the right latest. First commit keeps the per-declaration pairing and uses it for the bucket and the printed target, and routes `currentVersions`/`upToDateVersions` through the same multi-map as upgrade/downgrade so both rows survive to the reporter (per #65 / 486024f).

Second commit fixes a latent bug this exposed: `updateKey` deduplicated the multi-map keys by mutating them in place, and the result is rebuilt per output sink, so file reports collapsed rows that stdout printed.

When the same declared version has two different latests (the missing-repository case from #348), the greater one still wins.

### DIFF
--- a/gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/DependencyUpdates.kt
+++ b/gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/DependencyUpdates.kt
@@ -104,15 +104,11 @@ class DependencyUpdates
       unresolved: Set<UnresolvedDependency>,
       projectUrls: Map<Map<String, String>, String>,
     ): DependencyUpdatesReporter {
-      val currentVersions =
-        versions.current
-          .associateBy({ mapOf("group" to it.groupId, "name" to it.artifactId) }, { it })
+      val currentVersions = toMap(versions.current)
       val latestVersions =
         versions.latest
           .associateBy({ mapOf("group" to it.groupId, "name" to it.artifactId) }, { it })
-      val upToDateVersions =
-        versions.upToDate
-          .associateBy({ mapOf("group" to it.groupId, "name" to it.artifactId) }, { it })
+      val upToDateVersions = toMap(versions.upToDate)
       val downgradeVersions = toMap(versions.downgrade)
       val upgradeVersions = toMap(versions.upgrade)
 
@@ -123,7 +119,7 @@ class DependencyUpdates
         project, revision, outputFormatterArgument, outputDir,
         reportfileName, currentVersions, latestVersions, upToDateVersions, downgradeVersions,
         upgradeVersions, versions.undeclared, unresolved, projectUrls, gradleUpdateChecker,
-        gradleReleaseChannel,
+        gradleReleaseChannel, versions.latestByCurrent,
       )
     }
 
--- a/gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/DependencyUpdatesReporter.kt
+++ b/gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/DependencyUpdatesReporter.kt
@@ -188,26 +188,20 @@ class DependencyUpdatesReporter(
 
   private fun buildCurrentGroup(): MutableSet<Dependency> {
     return sortByGroupAndName(upToDateVersions)
-      .map { dep ->
-        updateKey(dep.key as HashMap)
-        buildDependency(dep.value, dep.key)
-      }.toSortedSet()
+      .map { dep -> buildDependency(dep.value, strippedKey(dep.key)) }
+      .toSortedSet()
   }
 
   private fun buildOutdatedGroup(): MutableSet<DependencyOutdated> {
     return sortByGroupAndName(upgradeVersions)
-      .map { dep ->
-        updateKey(dep.key as HashMap)
-        buildOutdatedDependency(dep.value, dep.key)
-      }.toSortedSet()
+      .map { dep -> buildOutdatedDependency(dep.value, strippedKey(dep.key)) }
+      .toSortedSet()
   }
 
   private fun buildExceededGroup(): MutableSet<DependencyLatest> {
     return sortByGroupAndName(downgradeVersions)
-      .map { dep ->
-        updateKey(dep.key as HashMap)
-        buildExceededDependency(dep.value, dep.key)
-      }.toSortedSet()
+      .map { dep -> buildExceededDependency(dep.value, strippedKey(dep.key)) }
+      .toSortedSet()
   }
 
   private fun buildUndeclaredGroup(): MutableSet<Dependency> {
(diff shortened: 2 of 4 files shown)


ben-manes:
thanks again!

Issue #910, opened 2024-10-30
Avoid using `Task.project` at execution time as that is being deprecated in Gradle 8.12


abstratt:
Avoid using `Task.project` at execution time as that is being deprecated in Gradle 8.12.

That happens in multiple places here:

https://github.com/ben-manes/gradle-versions-plugin/blob/516df34cd5a09014384af799a4d21870593c0207/gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/DependencyUpdatesTask.kt#L124

ben-manes:
I’m not actively working on this plugin. It was a weekend experiment in 2012 for learning Gradle (at 1.x then). If you or others wish to rewrite to avoid this issue then contributions are warmly welcome.

dtrunk90:
This plugin will become incompatible with Gradle >= 10:

> Invocation of Task.project at execution time has been deprecated. This will fail with an error in Gradle 10. This API is incompatible with the configuration cache, which will become the only mode supported by Gradle in a future release. Consult the upgrading guide for further information: https://docs.gradle.org/9.2.1/userguide/upgrading_version_7.html#task_project

alamothe:
Please fix @ben-manes

ianbrandt:
`Task.project` is no longer touched at execution time—the task action reads only its serialized inputs. This should have been addressed by PR #994, released in [v0.55.0](https://github.com/ben-manes/gradle-versions-plugin/releases/tag/v0.55.0). Feel free to reopen if you find otherwise.

Issue #367, opened 2019-12-30
Versions of plugins from settings.gradle are not checked


koral--:
Steps to reproduce:
- use Gradle 6.x
- add some plugin with old version to `settings.gradle(.kts)` for example "com.gradle.enterprise"
- invoke `updateDependencies` task

```
plugins {
    id("com.gradle.enterprise") version "3.1"
}
```
At the time of writing there is a version 3.1.1.

More info about build scan plugin: https://docs.gradle.com/enterprise/gradle-plugin/#gradle_6_x_and_later

ben-manes:
I suppose we would need to use [Settings](https://docs.gradle.org/current/javadoc/org/gradle/api/initialization/Settings.html) and evaluate both `getBuildscript()` and `getPluginManagement()`.

nedtwigg:
Closely related, it also does not check for these in the `settings.gradle` file:

```gradle
pluginManagement {
  plugins { ...
```
```

ben-manes:
Yes, see my comment above @nedtwigg. We would need to use both of those methods to capture those dependencies. Since those are not configurations, we'd have to make a detached configuration with the dependencies, repositories, and resolution strategy that those methods provide. Then resolve and print in the report with everything else.

jaredsburrows:
It looks like `dependabot` is able to do this: https://github.com/jaredsburrows/cs-interview-questions/pull/29.

ben-manes:
Yes, if I recall correctly they use simple regex parsing of the build files rather than executing it. That has other limitations.

The long-term approach that Github/Gradle are pursuing is outlined in this [proposal](https://docs.google.com/document/d/1TjxJJwgPavw-TFzK3110iH-CWstgdcVdb2JYiRy2GVs/edit) and [github issue](https://github.com/dependabot/dependabot-core/issues/1164). This would be similar to having static analyzers produce SARIF reports that the Github Security dashboard consumes, so it could simply be an action that feeds the current dependency graph into Github for a monitoring dashboard.

Vampire:
https://github.com/jmfayard/refreshVersions can also handle it, so would be nice to have it here too :-)

ben-manes:
PRs welcome. 😁

Vampire:
As always and everywhere :-D

ianbrandt:
Just adding a use case, the [Foojay Toolchains Plugin](https://github.com/gradle/foojay-toolchains#usage).

[...]

PR #994, opened 2026-07-21
Aggregate updates from per-project tasks (#948, #666, #910)

### LINKED ISSUE #666: Configuration Cache support
Got this warning in  Android Studio's build analyzer:

```
com.github.ben-manes.versions: not compatible 
You could save about 4.3s by turning Configuration cache  on. 
With Configuration cache, Gradle can skip the configuration phase entirely when nothing that affects the build configuration has changed.

The version of this plugin used in this build is not compatible with Configuration cache and we don’t know the version when it becomes compatible.  
 
 Plugin version: 0.42.0 
 Plugin dependency: com.github.ben-manes:gradle-versions-plugin
```

Is it possible to use configuration cache with this plugin?

### LINKED ISSUE #948: Only build plugin updates being reported with Gradle 9.0.0-rc-1
It seems with Gradle [9.0.0-rc-1](https://docs.gradle.org/9.0.0-rc-1/release-notes.html) only build plugin updates are being reported, not build dependency updates.

Gradle Versions Plugin 0.52.0, `gradlew dependencyUpdates`.

Gradle 8.14.2
https://github.com/sdkotlin/sd-kotlin-spring-talks/tree/d65c58a0efc608f71cb79cc8453ba5cf787c7772

```text
> Task :dependencyUpdates

 - com.autonomousapps.dependency-analysis:com.autonomousapps.dependency-analysis.gradle.plugin:2.18.0
 - com.fasterxml.jackson.module:jackson-module-kotlin:2.19.1
 - com.github.ben-manes.versions:com.github.ben-manes.versions.gradle.plugin:0.52.0
 - com.google.osdetector:com.google.osdetector.gradle.plugin:1.7.3
 - com.ninja-squad:springmockk:4.0.2
 - io.mockk:mockk:1.14.2
 - io.mockk:mockk-dsl:1.14.2
 - io.mockk:mockk-dsl-jvm:1.14.2
 - org.apache.logging.log4j:log4j-api-kotlin:1.5.0
 - org.assertj:assertj-core:3.27.3
 - org.gradlex.jvm-dependency-conflict-detection:org.gradlex.jvm-dependency-conflict-detection.gradle.plugin:2.4
 - org.gradlex.jvm-dependency-conflict-resolution:org.gradlex.jvm-dependency-conflict-resolution.gradle.plugin:2.4
 - org.jetbrains.kotlin:kotlin-allopen-compiler-plugin-embeddable:2.1.21
 - org.jetbrains.kotlin:kotlin-build-tools-impl:2.1.21
 - org.jetbrains.kotlin:kotlin-compiler-embeddable:2.1.21
 - org.jetbrains.kotlin:kotlin-klib-commonizer-embeddable:2.1.21
 - org.jetbrains.kotlin:kotlin-scripting-compiler-embeddable:2.1.21
 - org.jetbrains.kotlin:kotlin-stdlib:2.1.21
 - org.jetbrains.kotlin.jvm:org.jetbrains.kotlin.jvm.gradle.plugin:2.1.21
 - org.jetbrains.kotlinx:kotlinx-coroutines-core:1.10.2
 - org.jetbrains.kotlinx:kotlinx-coroutines-core-jvm:1.10.2
 - org.jetbrains.kotlinx:kotlinx-coroutines-reactor:1.10.2
 - org.jetbrains.kotlinx:kotlinx-coroutines-test:1.10.2
 - org.jetbrains.kotlinx:kotlinx-coroutines-test-jvm:1.10.2
 - org.jetbrains.kotlinx:kotlinx-datetime:0.6.2
 - org.jetbrains.kotlinx:kotlinx-datetime-jvm:0.6.2
 - org.junit.jupiter:junit-jupiter-api:5.13.1
 - org.junit.jupiter:junit-jupiter-params:5.13.1
 - org.sdkotlin.platforms:app-platform:1.0.0-SNAPSHOT
 - org.sdkotlin.platforms:test-platform:1.0.0-SNAPSHOT
 - org.springframework:spring-beans:6.2.8
 - org.springframework:spring-context:6.2.8
 - org.springframework:spring-core:6.2.8
 - org.springframework:spring-test:6.2.8
 - org.springframework:spring-web:6.2.8

 - com.github.ProjectMapK:jackson-module-kogera [2.19.0-beta24 -> 2.19.0-beta25]
     https://github.com/ProjectMapK/jackson-module-kogera
 - org.apache.logging.log4j:log4j-core [2.24.3 -> 2.25.0]
     https://logging.apache.org/log4j/2.x/
 - org.apache.logging.log4j:log4j-slf4j2-impl [2.24.3 -> 2.25.0]
     https://logging.apache.org/log4j/2.x/
 - org.springframework.boot:org.springframework.boot.gradle.plugin [3.5.0 -> 3.5.1]
     https://spring.io/projects/spring-boot
 - org.springframework.boot:spring-boot [3.5.0 -> 3.5.1]
     https://spring.io/projects/spring-boot
 - org.springframework.boot:spring-boot-autoconfigure [3.5.0 -> 3.5.1]
     https://spring.io/projects/spring-boot
 - org.springframework.boot:spring-boot-starter [3.5.0 -> 3.5.1]
     https://spring.io/projects/spring-boot
 - org.springframework.boot:spring-boot-starter-json [3.5.0 -> 3.5.1]
     https://spring.io/projects/spring-boot
 - org.springframework.boot:spring-boot-starter-log4j2 [3.5.0 -> 3.5.1]
     https://spring.io/projects/spring-boot
 - org.springframework.boot:spring-boot-starter-test [3.5.0 -> 3.5.1]
     https://spring.io/projects/spring-boot
 - org.springframework.boot:spring-boot-starter-webflux [3.5.0 -> 3.5.1]
     https://spring.io/projects/spring-boot
 - org.springframework.boot:spring-boot-test [3.5.0 -> 3.5.1]
     https://spring.io/projects/spring-boot
 - org.springframework.boot:spring-boot-test-autoconfigure [3.5.0 -> 3.5.1]
     https://spring.io/projects/spring-boot

Gradle current updates:
 - Gradle: [8.14.2: UP-TO-DATE]

Generated report file build/dependencyUpdates/report.txt
```

Gradle 9.0.0-rc-1
https://github.com/sdkotlin/sd-kotlin-spring-talks/tree/e012a8545feebea7733578525199df1b65bd8d04

```text
> Task :dependencyUpdates

 - com.autonomousapps.dependency-analysis:com.autonomousapps.dependency-analysis.gradle.plugin:2.18.0
 - com.github.ben-manes.versions:com.github.ben-manes.versions.gradle.plugin:0.52.0
 - com.google.osdetector:com.google.osdetector.gradle.plugin:1.7.3
 - org.gradlex.jvm-dependency-conflict-detection:org.gradlex.jvm-dependency-conflict-detection.gradle.plugin:2.4
 - org.gradlex.jvm-dependency-conflict-resolution:org.gradlex.jvm-dependency-conflict-resolution.gradle.plugin:2.4
 - org.jetbrains.kotlin.jvm:org.jetbrains.kotlin.jvm.gradle.plugin:2.1.21

 - org.springframework.boot:org.springframework.boot.gradle.plugin [3.5.0 -> 3.5.1]
     https://spring.io/projects/spring-boot

Gradle current updates:
 - Gradle: [9.0.0-rc-1: UP-TO-DATE]

Generated report file build/dependencyUpdates/report.txt
```

Diff: https://github.com/sdkotlin/sd-kotlin-spring-talks/commit/e012a8545feebea7733578525199df1b65bd8d04

ianbrandt:
Fixes #948. Fixes #666. Alternative for #983. Adds support for [isolated projects](https://docs.gradle.org/9.7.0-rc-1/release-notes.html#isolated-projects) tested against Gradle 9.7.0-rc-1.

Per-project tasks resolve their own configurations and a root task aggregates their results, jacoco-style, as suggested in #666. Unlike #983's configuration-time resolution, this stays fresh across a configuration cache hit and never reaches across projects. It replaces the existing resolution path rather than sitting beside it: the minimum Gradle version becomes 8.4, where the configuration-role APIs it uses exist, and the cross-project path is removed.

Each project that applies the plugin registers a `dependencyUpdatesPartial` task that resolves its own configurations and writes a serializable partial model, exposed through an outgoing variant (`Category.VERIFICATION`). The root `dependencyUpdates` task consumes the variants through an `ArtifactView` filtered to `ProjectComponentIdentifier`, then merges and reports as today. No project touches another project's configurations, so the Gradle 9 error in #948 has nothing to fire on, and `--parallel` works.

Freshness under the configuration cache: resolution is carried by a `Provider` realized after all projects are configured, right before the cache entry is stored, as @Vampire suggested in #983. Gradle fingerprints the metadata read during resolution, so publishing a new version invalidates the entry and the next run re-resolves. A cache hit means nothing changed, not a stale replay.

The reporters and `VersionMapping` take a project path, a `Logger`, and a resolved `File` instead of a `Project`, and `outputDir` defaults through `ProjectLayout`, which removes the execution-time `Task.project` usage from #910.

Preserved: task name, `revision`, `gradleReleaseChannel`, `outputDir`, `reportfileName` (and their `-D` overrides), `rejectVersionIf`, `resolutionStrategy(Action)`, all four `outputFormatter` forms, `filterConfigurations`, every report format, and the merged root report from root-only application. On three real multi-project builds the new report is byte-identical to the existing one, across runs and cache store/hit, with one deliberate exception called out below. A custom `outputFormatter` now works under the configuration cache. The reporter constructors that take a `Project` are deprecated rather than removed, so an `outputFormatter` closure that builds `PlainTextReporter(project, …)` still compiles and runs; the `Reporter` interface is untouched.

Isolated projects: the resolution and aggregation core is already clean under it; only cross-project task registration isn't. Applying the plugin per project works under isolated projects, and the root task warns about any project it received no partial result from, skipping projects with no build script of their own, which can't apply it. Keeping "apply once at the root" under isolated projects needs a settings plugin—a short stacked PR: https://github.com/ianbrandt/gradle-versions-plugin/pull/1. It's additive rather than an alternative: it applies this same project plugin to every project, the two surfaces can be applied together, and the report is identical either way. This PR stands on its own. The tradeoff in also taking the settings plugin is a second published plugin id, versus isolated projects having no apply-once story until it lands. If you want it, I'll merge the stacked PR into this branch and this PR will include it.

On the minimum Gradle version: the branch sets it to 8.4, where the aggregation APIs appeared. 8.8 is the other defensible floor, since it adds `gradle.lifecycle.beforeProject`, the isolated-projects-designated hook; the settings plugin deliberately uses the older `gradle.beforeProject` instead, so 8.8 would buy designated-API purity and nothing else. It's a one-constant change if you'd rather have it. Relatedly, the specs that pin Gradle 9 guard on JVM 17, so they run locally but skip on the current Java 8/11 CI matrix; running them on CI would take adding a newer Java version to the matrix.

One report change to call out for anyone parsing the JSON or XML: `unresolved[].reason` previously embedded the resolution failure's stack trace, which wasn't deterministic between two runs. It now carries the failure's message instead.


ben-manes:
Thanks for working through these issues!

> PlainTextReporter(project, …)

Can we add a deprecated constructor that forwards to `PlainTextReporter(project.path, …)`? This would be a more graceful update.

> minimum Gradle version

I think its fine increasing it where helpful as users on older versions can use the prior releases.

README.md
```diff
@@ -150,6 +150,22 @@ In a multi-project build, running this task in the root project will generate a
...
+which is not configuration-cache compatible.
+
+The resolution happens while the cache entry is stored, not while the task runs. Everything set on
+`dependencyUpdates`—`revision`, `checkConstraints`, `filterConfigurations`, `resolutionStrategy`
+and `rejectVersionIf`—has therefore been applied by the time the partial results are written, and
+a build that reuses the cache entry replays them.
+
+A report served from a reused entry is not stale. The dependency metadata read while resolving is
+tracked as a cache input, so publishing a new version of a dependency invalidates the entry and the
+next build resolves again.
```
ben-manes: I would lean towards dropping support for anything older than 8.4 (10-2023). You could then delete this doc update.

README.md
```diff
@@ -725,6 +741,24 @@ tasks.named("dependencyUpdates").configure {
...
+> ```groovy
+> def projectPath = project.path
+> tasks.named("dependencyUpdates").configure {
+>   outputFormatter = { result ->
+>     def reporter = new com.github.benmanes.gradle.versions.reporter.PlainTextReporter(
+>       projectPath, 'milestone', 'current')
+>     reporter.write(System.out, result)
+>   }
+> }
+> ```
```
ben-manes: this won't make any sense to casual users

gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/DependencyUpdatesTask.kt
```diff
@@ -37,7 +50,13 @@ open class DependencyUpdatesTask : DefaultTask() { // tasks can't be final
   /** Returns the outputDir destination. */
   @Input
   var outputDir: String =
-    "${project.buildDir.path.replace(project.projectDir.path + "/", "")}/dependencyUpdates"
+    "${
+      project.layout.buildDirectory
+        .get()
+        .asFile
+        .relativeTo(project.layout.projectDirectory.asFile)
```
ben-manes: 🤖 *Review note from Claude (via Claude Code), shared by Ben — not a human review comment.*

The `outputDir` default swapped `String.replace` for `File.relativeTo`, which regresses two edge cases the old code tolerated:

```kotlin
"${project.layout.buildDirectory.get().asFile
    .relativeTo(project.layout.projectDirectory.asFile).path}/dependencyUpdates"
```

- **Different filesystem roots throw.** Kotlin's `File.relativeTo` throws `IllegalArgumentException("this and base files have different roots")` when the build dir and project dir are on different roots — e.g. a Windows build with `layout.buildDirectory` redirected to another drive (`D:\…`) than the project (`C:\…`). It runs in the field initializer, so the task can't even be constructed and *every* `dependencyUpdates` invocation fails. The old `buildDir.path.replace(projectDir.path + "/", "")` left the absolute path untouched in that case.
- **Equal dirs → absolute path.** If the build dir equals the project dir, `relativeTo` returns an empty path, so the default becomes `"/dependencyUpdates"`, which `outputDirectory()` / `project.file()` then treat as absolute and write to the filesystem root.

Guarding the relativization (fall back to the absolute build path when the roots differ or the result is empty) restores the prior behavior. Not an issue for the common `<project>/build` layout, but it's a real regression for the cross-drive case.

gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/DependencyUpdatesTask.kt
```diff
@@ -97,32 +116,82 @@ open class DependencyUpdatesTask : DefaultTask() { // tasks can't be final
...
+  var projectPath: String = project.path
+
+  /** The project paths expected to contribute partial results, wired by the plugin. */
+  @Internal
+  var aggregatedProjectPaths: Set<String> = emptySet()
+
+  /** Captured at configuration time; replaces `project.file()` at execution. */
+  @get:Internal
+  val projectDirectory: DirectoryProperty =
+    project.objects.directoryProperty().convention(project.layout.projectDirectory)
```
ben-manes, line 176: 🤖 *Review note from Claude (via Claude Code), shared by Ben — not a human review comment.*

This branch drops `'5.0'` from `DifferentGradleVersionsSpec`, and this `projectDirectory` default uses `DirectoryProperty.convention()` — a Gradle **5.1** API — so Gradle 5.0 would now `NoSuchMethodError` at task construction. But `VersionsPlugin.requireMinimumGradleVersion()` still only rejects `< 5.0`, so the declared floor is untested and effectively broken.

Worth bumping the guard to match — 5.1 at minimum, or higher per the README note about dropping support below 8.4.

gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/DependencyUpdatesTask.kt
```diff
@@ -144,6 +213,40 @@ open class DependencyUpdatesTask : DefaultTask() { // tasks can't be final
...
+          "must apply the com.github.ben-manes.versions plugin to be aggregated when isolated " +
+          "projects is enabled, and projects that share a group and name are aggregated as one.",
+      )
+    }
+    val statuses =
+      mergeStatuses(partials.flatMap { it.statuses }) +
+        mergeStatuses(partials.flatMap { it.buildscriptStatuses })
+
+    DependencyUpdates.reporterFor(
+      statuses, projectPath, logger, revision, outputFormatter(), outputDirectory(), reportfileName,
```
ben-manes, line 204: 🤖 *Review note from Claude (via Claude Code), shared by Ben — not a human review comment.*

Edge case worth a look: each producer resolves its latest versions at the revision from its own ancestor chain (`service.resolve(project.path).revision` in `Aggregation.kt`), but `aggregateUpdates()` labels the whole report with the root accumulator's single `revision` here. So if a subproject sets `dependencyUpdates.revision = 'release'` while the root stays `'milestone'` (and no global `-Drevision`), that subproject's release-resolved versions get written under `available.milestone` in the JSON/XML report, and a consumer reading `available.release` sees nothing.

The legacy path never hit this because it resolved everything with one revision. Likely fine to leave since per-project `revision` isn't really a documented feature — flagging in case you'd prefer the producers and the report to agree on one revision.

README.md
ianbrandt: Sounds good. Done.

README.md
ianbrandt: Removed.

gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/DependencyUpdatesTask.kt
ianbrandt: Fixed. The default keeps the absolute path when relativizing would throw (different roots) or read as the filesystem root (build directory equal to the project directory), matching the old `String.replace` behavior. Regression test added for the equal-directories case.

gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/DependencyUpdatesTask.kt
ianbrandt: Done.

ianbrandt:
> Thanks for working through these issues!

Thanks for the quick review!

> > PlainTextReporter(project, …)
> 
> Can we add a deprecated constructor that forwards to `PlainTextReporter(project.path, …)`? This would be a more graceful update.

Done. Added for all four reporters.

> > minimum Gradle version
> 
> I think its fine increasing it where helpful as users on older versions can use the prior releases.

Changed to 8.4 and dropped the `-Dcom.github.benmanes.versions.aggregate=false` backwards compatibility support.

gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/DependencyUpdatesTask.kt
ianbrandt: I left this as is. The report format carries a single revision, so the options are the root's label with the subproject's resolution (current), or ignoring the subproject's setting. Running `dependencyUpdates` on the subproject labels and resolves consistently. I could pin the producers to the accumulator's `revision` if you'd rather they agree.

ianbrandt:
I found two issues testing locally and just pushed fixes:
* An aggregate invoked by path under `--configure-on-demand` collected nothing from the other projects and reported the root alone.
* A container project with no build script of its own was named in the missing-projects warning on every build even though there is nothing it can do about it.

Both have specs.

ben-manes (approved):

ben-manes:
Thanks! Whenever you're ready to release all of your fixes, you can bump the version number and create a Github Release. That will kickoff the github action to build and publish the plugin.

ianbrandt:
Thanks, @ben-manes!

I might have buried the lead on it above, and this may be my fault for marking this branch ready instead of keeping it a draft until we discussed this, but let me know if you're interested in https://github.com/ianbrandt/gradle-versions-plugin/pull/1. Personally, I'm partial to the explicit declaration approach, but both can be supported as alternatives if you want to take on publishing the additional plugin. I could repoint that PR at master.

ben-manes:
One quirk of settings plugins is that their version is currently not checked. It might be possible but I never looked into it much (see https://github.com/ben-manes/gradle-versions-plugin/issues/367). Otherwise I am fine if you want to give it a go

ianbrandt:
@ben-manes, Thanks for the collaborator invite. I'm looking forward to contributing to this awesome plugin as much as I can.

Regarding release and the settings plugin possibility, I've been exploring that and testing the changes from this PR further. I've realized something I don't like about these changes. As is, when enabling `--isolated-projects`, not only do you have to apply the plugin directly to every project you want checked for dependency updates (e.g., via a convention script or plugin), you also have to switch to invoking it by path with `:dependencyUpdates` rather than by name with `dependencyUpdates`. A one-time documented migration step for build authors seems to me like an okay ask, but changing the way the plugin is typically invoked for build users is another thing. I believe I have an approach to addressing this by way of a split aggregator vs. contributor plugin pattern. I'm proving it out now, and will submit a follow-up PR shortly if successful. I'd suggest holding off on the release or reverting this PR until I do. I should have something today or tomorrow at the latest.

I believe the updated approach precludes the settings plugin option, so it'd resolve that question (favorably, I'd say).

ben-manes:
Sure, sounds good! I'll defer to you to publish. I can give it a whirl on some of my own projects too. The test suite passing and the clarity of the code changes made me feel fairly good about it, but then I haven't gotten `--isolated-projects` working elsewhere yet so I suspect most others haven't either as its still early days.

ianbrandt:
Follow-up from testing this against real builds under isolated projects: #995 adds a producer-only plugin id that keeps bare `dependencyUpdates` working unchanged, instead of the settings plugin. Given that, I'm inclined to drop the settings plugin idea unless someone asks for apply-once; it also sidesteps the version-check quirk from #367 entirely, since the contributor id is an ordinary project plugin on the build classpath.

README.md
ben-manes: I was probably wrong and should be added back in, since I ran into it 🙂

PR #995, opened 2026-07-22
Add a contributor plugin that preserves named `dependencyUpdates` invocation under isolated projects

ianbrandt:
Follow-up to #994 from testing it against real builds under isolated projects. Isolated projects needs every project to apply the plugin itself, but applying `com.github.ben-manes.versions` everywhere puts a `dependencyUpdates` task in every project, so running `dependencyUpdates` by name produces a report per project instead of the single merged report people run today.

Adds `com.github.ben-manes.versions.contributor`, a producer-only project plugin that registers just the partial task and its outgoing variant. Under isolated projects the root keeps `com.github.ben-manes.versions` and every other project applies the contributor id, typically from a shared convention plugin. `dependencyUpdates` then matches only the root task, so build users invoke it exactly as before and get the single merged report.

The main plugin is unchanged and is a superset of the contributor plugin: applying it per project still creates a per-project `dependencyUpdates` task for separate reports and still feeds the aggregate, so no project needs both. Also renames the partial task to `partialDependencyUpdates`: in a project with only a producer, task abbreviation would otherwise match `dependencyUpdates` to the old name and silently succeed with no report. A regression test covers it. Documents the setup in a new "Isolated projects" README section, along with the aggregation compatibility notes and the settings-inheritance behavior from #994 that weren't written up yet.

Verified against three real multi-project builds (19, 12, and 45 projects) with isolated projects, configuration cache, configure on demand, and parallel build all on: one merged report from bare `dependencyUpdates`, configuration cache reused on the second run.

When existing projects without isolated projects enabled take up the new version of this plugin (i.e. only applying `com.github.ben-manes.versions` at the root), they continue to work as they did before, except now with configuration cache and parallel build support.

Prior art for the split: Gradle's own [jacoco-report-aggregation](https://docs.gradle.org/current/userguide/jacoco_report_aggregation_plugin.html) and [test-report-aggregation](https://docs.gradle.org/current/userguide/test_report_aggregation_plugin.html) aggregate the same way—one project resolves `Category.VERIFICATION` variants that the others expose, which is the mechanism #994 already mirrors. The [Baseline Profile plugin](https://developer.android.com/topic/performance/baselineprofiles/create-baselineprofile) ships explicit `androidx.baselineprofile.producer`/`.consumer` ids for the same producer/consumer split, and the [Dependency Analysis Gradle Plugin](https://github.com/autonomousapps/dependency-analysis-gradle-plugin) puts the aggregate `buildHealth` task at the root while `projectHealth` is the per-project surface.

This doesn't preclude the settings plugin floated on #994: if apply-once demand shows up, one can sit on top of this, applying the contributor plugin to every project and the main plugin to the root, with the same invocation. I've closed ianbrandt/gradle-versions-plugin#1 unmerged accordingly—as written it applied the main plugin everywhere, which recreates the invocation change this PR avoids.

The plugin id is provisional—happy to rename.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ checks for updates to Gradle itself.
   - [Related plugins](#related-plugins)
 - [dependencyUpdates](#dependencyupdates)
   - [Multi-project build](#multi-project-build)
+  - [Isolated projects](#isolated-projects)
   - [Revisions](#revisions)
   - [RejectVersionsIf and componentSelection](#rejectversionsif-and-componentselection)
   - [Gradle Release Channel](#gradle-release-channel)
@@ -144,6 +145,39 @@ field explaining failures or missing information. The update check may be disabl
 In a multi-project build, running this task in the root project will generate a consolidated/merged
 report for dependency updates in all subprojects. Alternatively, you can run the task separately in
 each subproject to generate separate reports for each subproject.
+
+The report is aggregated from a task in each project, so it is compatible with parallel execution
+and the configuration cache. Each project takes the settings that control resolution (`revision`,
+`rejectVersionIf` or a full `resolutionStrategy`, `filterConfigurations`, `checkConstraints`, and
+`checkBuildEnvironmentConstraints`) from the nearest project up the hierarchy whose task set them,
+so configuring the root project's task covers every project unless a subproject configures its own.
+
+#### Isolated projects
+
+Under [isolated projects](https://docs.gradle.org/current/userguide/isolated_projects.html) a
+project plugin cannot register a task in another project, so a project only contributes to the
+aggregate report if it applies a plugin itself. Keep applying `com.github.ben-manes.versions` in
+the root project, and apply `com.github.ben-manes.versions.contributor` in every other project,
+typically from a convention plugin they already share:
+
+```kotlin
+// buildSrc/src/main/kotlin/my-conventions.gradle.kts
+plugins {
+  id("com.github.ben-manes.versions.contributor")
+}
+```
+
+The contributor plugin registers only the task that feeds the aggregate report, so
+`dependencyUpdates` remains a single task in the root project. The main plugin is a superset of
+the contributor plugin: a project that applies `com.github.ben-manes.versions` instead still feeds
+the aggregate report, and also gets its own `dependencyUpdates` task covering itself and its
+subprojects. So for a separate per-project report, apply the main plugin to that project—there is
+no need to apply both. Running `dependencyUpdates` by name then includes that project's report,
+as it always has when the plugin is applied to more than one project; run a single task by its
+path to get just that report (e.g., `:subproject:dependencyUpdates`).
+
+Under isolated projects, contributing projects that share a group and name are aggregated as one;
+the console warns about any project the report is missing.
 
 #### Revisions
 
@@ -371,8 +405,8 @@ Have a look at [`examples/groovy`](https://github.com/ben-manes/gradle-versions-
 $ ./gradlew publishToMavenLocal
 
 # Try out the samples
-$ ./gradlew -p examples/groovy dependencyUpdate
-$ ./gradlew -p examples/kotlin dependencyUpdate
+$ ./gradlew -p examples/groovy dependencyUpdates
+$ ./gradlew -p examples/kotlin dependencyUpdates
 ```
 
 ### Report format
--- a/README.md
+++ b/README.md
@@ -167,7 +167,11 @@ plugins {
 }
 ```
 
-The contributor plugin registers only the task that feeds the aggregate report, so
+The convention plugin's own build must have the plugin on its classpath, e.g. as an
+`implementation("com.github.ben-manes:gradle-versions-plugin:$version")` dependency in
+`buildSrc/build.gradle.kts`.
+
+The contributor plugin registers only the producer that feeds the aggregate report, so
 `dependencyUpdates` remains a single task in the root project. The main plugin is a superset of
 the contributor plugin: a project that applies `com.github.ben-manes.versions` instead still feeds
 the aggregate report, and also gets its own `dependencyUpdates` task covering itself and its
--- a/gradle-versions-plugin/build.gradle.kts
+++ b/gradle-versions-plugin/build.gradle.kts
@@ -53,5 +53,12 @@ gradlePlugin {
       description = properties["POM_DESCRIPTION"].toString()
       tags.set(listOf("dependencies", "versions", "updates"))
     }
+    create("versionsContributorPlugin") {
+      id = properties["PLUGIN_CONTRIBUTOR_NAME"].toString()
+      implementationClass = properties["PLUGIN_CONTRIBUTOR_NAME_CLASS"].toString()
+      displayName = properties["POM_CONTRIBUTOR_NAME"].toString()
+      description = properties["POM_CONTRIBUTOR_DESCRIPTION"].toString()
+      tags.set(listOf("dependencies", "versions", "updates"))
+    }
   }
 }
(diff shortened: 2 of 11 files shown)


ben-manes (approved):
lgtm, ran Claude's pr-review skill so posting its findings as optional feedback

ben-manes:
## Review by Claude Code 🤖

> _This review was generated by **Claude Code** (Opus 4.8) and posted on the maintainer's behalf. It's an automated, multi-aspect pass — code quality, test coverage, error handling, type design, docs/comments, and simplification. Claims were checked against the code, and the load-bearing ones were verified empirically (noted below); please still confirm before acting._

**Verdict: solid, mergeable PR — no critical issues and no correctness bugs.** The core mechanism was verified sound across every plugin-combination/ordering permutation, the central fix was *empirically* confirmed against Gradle's real task matcher, and the branch compiles cleanly. What follows is one recommended message tweak, some test-hardening, and optional polish.

### ✅ Independently verified (facts, not assertions)
- **The task rename genuinely closes the silent-failure hole.** Running Gradle's actual `NameMatcher`: a bare `dependencyUpdates` invocation *does* silently select the old `dependencyUpdatesPartial` (→ `SKIPPED`, `BUILD SUCCESSFUL`, no report), while the new `partialDependencyUpdates` makes it **fail loudly**. No residual collision — the other `dependencyUpdates*` names are configurations/services, not invocable tasks.
- **`registerProducer` idempotency holds** across all cases (main+contributor, both application orders, isolated & non-isolated, root & leaf) → each project yields exactly one partial. The `count(...) == 1` test confirms it.
- **No new silent failures** — every swallow in the touched files predates this PR (traces to #994).
- **Builds clean** — main Kotlin + test Groovy compile; the `private → internal` widening is module-scoped and sound.
- **Docs are unusually accurate** — settings-inheritance, "superset / no need to apply both," and the `:subproject:dependencyUpdates` path example all check out against code and tests.

### 🔴 Critical (must fix before merge)
None.

### 🟠 Important (should address)

**1. The "report is missing" warning steers users toward the anti-pattern this PR exists to prevent** — `DependencyUpdatesTask.kt:194-196`
*Surfaced independently by multiple review passes.* The warning says a missing project "**must apply the `com.github.ben-manes.versions`** plugin … when isolated projects is enabled." But this PR establishes `com.github.ben-manes.versions.contributor` as the right plugin for subprojects — and `README.md:180` explicitly routes users to this very warning. A user who follows its literal advice applies the main plugin everywhere, re-creating the per-project `dependencyUpdates` proliferation the PR removes. The line is pre-existing (from #994) but *this PR makes it stale*. Fix is one line — name both plugins ("…must apply the `com.github.ben-manes.versions` **or `com.github.ben-manes.versions.contributor`** plugin…"). **Low risk:** tests only assert the `'The dependency updates report is missing'` prefix, so the tail is free to change.

**2. The config-cache reuse test under-asserts** — `ContributorAggregationSpec.groovy:160-170`
The *"Reuses the configuration cache across runs"* test checks only `outcome == SUCCESS` and `contains('Configuration cache entry reused.')`. It never asserts the aggregated **content survived** the cache hit, nor that **no false "report is missing" warning** was emitted — both of which every sibling cache test asserts (`AggregationSpec.groovy:107-111`, `KotlinMultiplatformAggregationSpec.groovy:96`). Config-cache serialization of contributor producers under isolated projects is a *new* surface and exactly where a "reused entry drops partials" regression would hide. Add the guice/guava content lines and `!contains('The dependency updates report is missing')` to the second-run `then:` block.

### 🟡 Suggestions

**3. The task-rename regression is proven at the root only** — `ContributorAggregationSpec.groovy:172-193`. The realistic case is a producer-only *subproject* (`./gradlew :app:dependencyUpdates`), which is untested; and `"Cannot locate tasks"` would also pass if the producer were never registered at all (false-pass). Add a subproject invocation and pair it with a positive assertion that `:app:partialDependencyUpdates` still runs, to disambiguate the two regressions.

**4. Extract the duplicated shared-service construction** — `VersionsContributorPlugin.kt:23-25` copies `Aggregation.kt:106-108` verbatim. A single internal helper (or a `registerProducer(project)` overload that acquires the service itself) would let `PARAMETERS_SERVICE` revert to `private`, drop two imports from the new plugin, turn the "superset" relationship into real composition, and encapsulate the correctness-critical "same service instance" invariant by construction. Behavior-preserving; author's-judgment call.

**5. Pin a stable Gradle, not the RC** — `ContributorAggregationSpec` hardcodes `9.7.0-rc-1` throughout (siblings use stable `9.0.0`/`9.6.1`) and asserts the version-specific string `'Configuration cache entry reused.'`. RCs get delisted from `services.gradle.org`, which would make the spec unreproducible. Move to `9.7.0` final once released.

**6. No isolated-projects test for the "share a group and name" README claim** — the only shared-group/name test (`AggregationSpec.groovy`) runs *non-isolated*, where the else-branch task-output wiring protects against eviction; the isolated path (`Aggregation.kt:163-177`) deliberately omits that wiring, so it's a genuinely different code path. Covering it would also close the positive missing-warning path in the contributor context.

### ⚪ Minor / follow-ups (mostly pre-existing — not blockers)
- **`statusesOf` swallow** (`Aggregation.kt:242-260`, pre-existing) drops unresolvable configurations at `info` level — invisible without `--info` — and this PR makes that path newly reachable in contributor-only subprojects. A sensible **separate follow-up**: raise to `warn` and narrow the broad `catch (Exception)` so it can't also mask logic bugs.
- **`isIsolatedProjectsEnabled`** (`Aggregation.kt:31-34`, pre-existing) swallows with no log line; a future Gradle changing `StartParameterInternal` would silently flip aggregation branches. Log the throwable at `debug`.
- **README convention-plugin snippet** (`README.md:163-168`) omits the `buildSrc` classpath prerequisite — copy-pasting without adding the plugin as a `buildSrc` `implementation` dependency yields "Plugin not found." A one-line note would help.
- **README phrasing nits**: "registers only the task" (`README.md:170`) also registers a consumable variant that carries the output; `POM_CONTRIBUTOR_DESCRIPTION`'s "root report" is slightly loose since the aggregator can live on a subproject.
- **Plugin-id literal** is duplicated between `gradle.properties` and Kotlin — a co-located `const val ID` would single-source it within the code (runtime derivation isn't cleanly available in Gradle).
- **SAX-parser guard asymmetry** (pre-existing): a contributor-only project on Gradle 8.4–8.10.2 resolves without the friendly SAX error the main plugin's `doFirst` provides.

### 🏆 Strengths
Genuinely excellent KDoc/comments (the `PARTIAL_TASK_NAME` rationale is empirically true), a strong settings-inheritance test (rejecting `3.1` at the *root* task and proving `3.0` reaches the contributor-only `:app` — verifiable no other way), a correct `count(...) == 1` idempotency assertion, comprehensive README, and a well-scoped new spec that targets the contributor *deltas* rather than re-testing #994's mechanism.

### Recommended action
1. **Do now (this PR):** update the warning message (**#1**) and strengthen the config-cache test (**#2**) — both small and low-risk.
2. **Nice to include:** the subproject rename test (**#3**) and stable-Gradle pin (**#5**).
3. **Optional / author's call:** the service-construction refactor (**#4**).
4. **Separate follow-up issue:** the `statusesOf` info-level swallow.

ianbrandt:
Thanks!

Addressed:
* (1) The missing-project warning now names both plugin ids.
* (2) The config-cache test asserts the content and the absence of a false missing warning across the reuse.
* (3) The rename regression now covers a producer-only subproject paired with a positive partial-task check.
* (4) The contributor plugin acquires the parameters service through a shared helper so the service key is private again.
* (6) New isolated-projects test pins the shared group-and-name behavior.

On the Gradle pin (5), I kept 9.7.0-rc-1 since it's the only currently available version with isolated projects incubating instead of experimental. I'll submit an update PR for 9.7.0 when it's final.

PR #998, opened 2026-07-22
Publish under the io.github.ben-manes namespace

### LINKED ISSUE #997: v0.55.0 failed to publish: portal rejects the new plugin id under a com.github group
The 0.55.0 release never reached the Plugin Portal. The [deploy run](https://github.com/ben-manes/gradle-versions-plugin/actions/runs/29944033006) was rejected with:

> Group id must not start with 'com.github', it should be replaced with 'io.github'

See ["The GitHub group ID prefix can't be com.github"](https://plugins.gradle.org/docs/publish-plugin#publication-failed).

The portal grandfathers existing `com.github` plugin ids (0.54.0 published fine in April, and com.github.spotbugs published on 2026-07-13) but applies the io.github policy to new ids. 0.55.0 introduces a new id, `com.github.ben-manes.versions.contributor` (#995), and `publishPlugins` publishes both plugins as one unit, so the new id's rejection failed the whole release. Same shape as gradle/gradle#19528.

It looks like the options may be:

1. Update just the new plugin to `io.github.ben-manes.versions.contributor` (see also https://github.com/gradle/gradle/issues/19528#issuecomment-1009701797)
2. Update both: `com.github.ben-manes.versions` -> `io.github.ben-manes.versions` and `com.github.ben-manes.versions.contributor` -> `io.github.ben-manes.versions.contributor`
3. Apply for an exception: https://github.com/gradle/plugin-portal-requests/issues

The Kotlin package can stay `com.github.benmanes.*`, since a package is independent of the Maven group.

I'd vote for (2). I can update the release notes to call out the breaking change, and then it's consistent and done and over with going forward. Option (1) is inconsistent group IDs, which may be confusing. Option (3) may take a while.

@ben-manes? @jaredsburrows?

I've prepared a PR for option (2) in case that ends up being the way we go: #998.

One prerequisite outside the code: io.github.ben-manes has to be approved on the portal account (verified against the GitHub account of the same name) before a release under it can succeed.

ianbrandt:
Fixes #997.

Moves the group id and both plugin ids from `com.github.ben-manes` to `io.github.ben-manes`, since the portal rejects new plugin ids under `com.github`—which is what failed the 0.55.0 release (details in the issue).

Per the discussion in #997, the old `com.github.ben-manes.versions` id stays published: a third plugin declaration backed by a thin wrapper class that logs a deprecation warning and applies `VersionsPlugin`. Its marker keeps its grandfathered com.github coordinates and depends on the `io.github.ben-manes` artifact, so existing builds keep resolving new releases and the rename is non-breaking for them (the com.diffplug.gradle.spotless playbook). The contributor plugin is io.github-only, since its com.github id is the one the portal rejected and it has no published users.

Deliberately unchanged:

- The Kotlin package `com.github.benmanes.gradle.versions`: a Java package is independent of the Maven group id, and renaming it would be a large source break for no portal benefit.
- The dogfooded plugin id in `gradle/libs.versions.toml`: it can move to the new id once a release under it exists on the portal.
- The `com.github.ben-manes:unresolvable` entries in the examples, specs, and README samples, which are deliberately unresolvable placeholders.

One prerequisite outside the code: publish with a Plugin Portal account linked to the ben-manes GitHub account—that is how the portal verifies ownership of `io.github.*` ids. There is no pre-approval UI: the first version of each new id goes through the portal's manual review after publishing (the grandfathered id skips it), so the new ids may take a while to appear on the first release.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -55,9 +55,9 @@ jobs:
       - name: Run examples
         run: |
           export VERSION_NAME=$(cat gradle.properties | grep -w "VERSION_NAME" | cut -d'=' -f2)
-          rm -rf ~/.m2/repository/com/github/ben-manes/gradle-versions-plugin/$VERSION_NAME
+          rm -rf ~/.m2/repository/io/github/ben-manes/gradle-versions-plugin/$VERSION_NAME
           ./gradlew publishToMavenLocal -x test -s
-          cat ~/.m2/repository/com/github/ben-manes/gradle-versions-plugin/$VERSION_NAME/gradle-versions-plugin-$VERSION_NAME.pom
+          cat ~/.m2/repository/io/github/ben-manes/gradle-versions-plugin/$VERSION_NAME/gradle-versions-plugin-$VERSION_NAME.pom
           cd examples/groovy/ && ../../gradlew dU && cd -
           cd examples/kotlin/ && ../../gradlew dU && cd -
 
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -3,6 +3,7 @@ name: deploy
 on:
   release:
     types: [ created ]
+  workflow_dispatch:
 
 jobs:
   publish:
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 [![Build](https://github.com/ben-manes/gradle-versions-plugin/workflows/build/badge.svg)](https://github.com/ben-manes/gradle-versions-plugin/actions)
-[![gradlePluginPortal](https://img.shields.io/maven-metadata/v/https/plugins.gradle.org/m2/com/github/ben-manes/versions/com.github.ben-manes.versions.gradle.plugin/maven-metadata.xml.svg?label=gradlePluginPortal)](https://plugins.gradle.org/plugin/com.github.ben-manes.versions)
+[![gradlePluginPortal](https://img.shields.io/maven-metadata/v/https/plugins.gradle.org/m2/io/github/ben-manes/versions/io.github.ben-manes.versions.gradle.plugin/maven-metadata.xml.svg?label=gradlePluginPortal)](https://plugins.gradle.org/plugin/io.github.ben-manes.versions)
 
 # Gradle Versions Plugin
 
@@ -35,14 +35,14 @@ You can add this plugin to your top-level build script using the following confi
 
 ```groovy
 plugins {
-  id "com.github.ben-manes.versions" version "$version"
+  id "io.github.ben-manes.versions" version "$version"
 }
 ```
 or via the
 
 ### `buildscript` block:
 ```groovy
-apply plugin: "com.github.ben-manes.versions"
+apply plugin: "io.github.ben-manes.versions"
 
 buildscript {
   repositories {
@@ -50,7 +50,7 @@ buildscript {
   }
 
   dependencies {
-    classpath "com.github.ben-manes:gradle-versions-plugin:$version"
+    classpath "io.github.ben-manes:gradle-versions-plugin:$version"
   }
 }
 ```
@@ -69,7 +69,7 @@ initscript {
   }
 
   dependencies {
-    classpath 'com.github.ben-manes:gradle-versions-plugin:+'
+    classpath 'io.github.ben-manes:gradle-versions-plugin:+'
   }
 }
 
@@ -96,7 +96,7 @@ initscript {
     gradlePluginPortal()
   }
   dependencies {
-    classpath("com.github.ben-manes:gradle-versions-plugin:+")
+    classpath("io.github.ben-manes:gradle-versions-plugin:+")
   }
 }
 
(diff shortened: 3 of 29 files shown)


ben-manes:
> 🤖 **Review by [Claude Code](https://claude.com/claude-code), posted by @ben-manes on his behalf.** Findings were cross-checked by three independent review passes (code, tests, docs) plus a manual trace; the notes below distinguish real items from things that were checked and cleared.

Thanks for the very thorough PR and writeup, @ianbrandt — the reasoning in the description made this easy to follow. Overall this is **solid: no correctness defects, and CI is green on JDK 8 and 11.** The spotless-playbook approach (grandfathered marker → new impl) is the right call, the migration is internally consistent (exactly one `com.github.ben-manes.versions` literal remains outside the shim and its spec), and the new behavior is genuinely exercised by `LegacyPluginIdSpec` rather than stubbed.

A few suggestions, none blocking.

## Worth changing before merge

**1. The Usage note over-promises for `buildscript` / `initscript` classpath consumers.** `README.md:31-33`

> That id is deprecated but keeps receiving releases, so existing builds continue to see updates

That's true for **`plugins {}`-block** users — the republished marker `com.github.ben-manes.versions:…gradle.plugin` repoints to the new `io.github.ben-manes:gradle-versions-plugin` impl, so they're covered. It's **not** true for users who reference the raw coordinate directly:

```groovy
buildscript { dependencies { classpath "com.github.ben-manes:gradle-versions-plugin:<v>" } }
```

That coordinate moved with the `GROUP` rename and receives no further releases, so those builds silently freeze at the last `com.github` release — and they never even see the new deprecation warning, since it lives in `LegacyVersionsPlugin` inside a jar they can no longer resolve at the old coordinate. Notably, `README.md:57` and `README.md:76` update exactly this snippet to `io.github…`, which helps *new* readers but leaves *existing* copy-pasters uncovered. The code move itself is unavoidable; this is a doc-completeness gap. Suggest one added sentence to the note, e.g.:

> If you apply the plugin via a `buildscript`/`initscript` `classpath` dependency rather than the `plugins {}` block, update the coordinate to `io.github.ben-manes:gradle-versions-plugin` — the old `com.github.ben-manes:gradle-versions-plugin` coordinate receives no further releases.

## Nice to have

**2. Add a negative test: the current `io.github` id must stay silent.** `LegacyPluginIdSpec.groovy`

The shim's contract is symmetric — warn on the old id, *stay silent* on the new one — but only the first half is asserted. Today that's correct (the `warn` lives only in `LegacyVersionsPlugin`), so this is pure regression insurance: it's nearly free and it guards a plausible future refactor ("hoist the shared logic into `VersionsPlugin`") that would otherwise leak the deprecation warning to **every** user of the non-deprecated id. A third feature method mirroring the existing setup:

```groovy
def 'current io.github id does not warn'() {
  given:
  testProjectDir.newFile('build.gradle') <<
    """
      buildscript {
        dependencies {
          classpath files($classpathString)
        }
      }

      apply plugin: 'io.github.ben-manes.versions'
    """.stripIndent()

  when:
  def result = GradleRunner.create()
    .withProjectDir(testProjectDir.root)
    .withArguments('dependencyUpdates')
    .build()

  then:
  !result.output.contains('is deprecated')
}
```

**3. The legacy plugin reuses the current plugin's portal `displayName`.** `build.gradle.kts:59`

Both `versionsPlugin` and `legacyVersionsPlugin` set `displayName = properties["POM_NAME"]` ("Gradle Versions Plugin"), so the two Plugin Portal listings render with the same title and are distinguishable only by the description. Since you already gave the legacy declaration a distinct `POM_LEGACY_DESCRIPTION`, a distinct display name would finish the thought and surface the deprecation in search/title — e.g. a `POM_LEGACY_NAME=Gradle Versions Plugin (deprecated id)` property wired into the legacy `displayName`.

## Optional polish

- **Per-apply `logger.warn`** (`LegacyVersionsPlugin.kt:15-17`) fires once per project under `allprojects { apply … }`. Acceptable — the main plugin is root-only by design and a deprecation warning is meant to be seen — but if the noise ever bites, it could be deduped to once per build via a build-scoped flag/service.
- **`LegacyVersionsPlugin` KDoc** describes the *why* and *what* well but not the two runtime behaviors; a trailing "Applying it logs a deprecation warning and delegates to [VersionsPlugin]." would make it self-contained.

## Checked and cleared (not issues — noting so they don't get re-raised)

- **Double `requireMinimumGradleVersion`** (legacy id, then `io.github` id inside `VersionsPlugin`): a pure guard with no side effects; on supported Gradle both are no-ops, and running the legacy call first means a user on Gradle < 8.4 sees the id they *actually applied* echoed in the error. Deliberate and correct.
- **Applying both ids**: `pluginManager.apply(VersionsPlugin::class.java)` dedupes by class, plus the `if (!tasks.names.contains("dependencyUpdates"))` guard — no double task/producer registration. Covered by the "both ids harmless" spec.
- **`KotlinDependencyUpdatesSpec` pinning `id 'io.github.ben-manes.versions' version '0.51.0'`** (a version never published under the new id): resolves from the injected `withPluginClasspath()`, which matches by id and ignores the version — verified empirically by running the spec on this branch (passes; the coordinate 404s on the portal, so a pass is only possible via the local classpath). Pre-existing: the `version '0.51.0'` pin is effectively dead either way, but that's unchanged from `master`.
- **Contributor plugin is `io.github`-only**: correct — it was added in b7482cd and only ever tagged in the failed 0.55.0 release, so it has no published users needing a grandfathered id.
- **Deliberately unchanged**: Kotlin package `com.github.benmanes.gradle.versions`, the dogfooded `libs.versions.toml` id, and the `com.github.ben-manes:unresolvable` sample-report data — all correctly left alone.

## Strengths

- Correct, minimal shim; the migration sweep is complete and consistent.
- `LegacyPluginIdSpec` covers the two behaviors that matter (old id warns + works; both ids co-applied is harmless), and the warning string is asserted verbatim consistently across the shim, the spec, and the README.
- Docs are factually accurate; the badge URL transform is right in both path segments and will go live after the first `io.github` release.

*Prereq restated for the release, per your description: the `io.github.ben-manes` namespace needs approval on the Plugin Portal account before the next publish, and `./gradlew publishPlugins --validate-only` can dry-run that — including that the grandfathered id validates alongside the new ones — without publishing.*

---
<sub>🤖 Generated by Claude Code (pr-review-toolkit) at @ben-manes's request. A human should still give it a pass.</sub>

ianbrandt:
Adopted all three: the migration note now covers `buildscript`/`initscript` classpath consumers, the deprecated listing gets a distinct display name, and the spec now asserts the current id stays silent (plus the KDoc line). Skipped the warn dedup—once per project seems like the right visibility for a deprecation, and it keeps the shim trivial.

ianbrandt:
Also added a `workflow_dispatch` trigger to deploy (ff40283), so we can retry the v0.55.0 publish against master directly instead of re-firing the release event, and move the tag once a publish actually sticks.

ben-manes:
oops, pushed ready for review by mistake. approved so you can merge whenever.

ben-manes (approved):

PR #999, opened 2026-07-23
Declare configuration cache compatibility on the Plugin Portal

ianbrandt:
Publishing 0.55.0 logged: "Consider declaring compatibility of your plugin with the following Gradle features: Configuration cache."

This bumps plugin-publish to 2.1.1 (the `compatibility` DSL needs 2.1.0+) and declares `configurationCache = true` on all three plugin ids, which lands in each plugin descriptor as `compatibility.feature.configuration-cache=DECLARED_SUPPORTED`. The portal reads that at publish time and shows a compatibility badge on the plugin page; it also warns that publishing without the declaration is deprecated.

### DIFF
--- a/gradle-versions-plugin/build.gradle.kts
+++ b/gradle-versions-plugin/build.gradle.kts
@@ -1,3 +1,5 @@
+import org.gradle.plugin.compatibility.compatibility
+
 plugins {
   alias(libs.plugins.kotlin.jvm)
   alias(libs.plugins.dokka)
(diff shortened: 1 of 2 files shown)


ben-manes (approved):

ianbrandt:
Unless we want to delete and republish v0.55.0 (it's publish is still pending approval), this change won't take effect until the next release (presumably v0.56.0).

Issue #727, opened 2023-02-10
Constraints checking and latest version determination issues with platforms


ianbrandt:
I'm defining my own [platform](https://docs.gradle.org/current/userguide/platforms.html#sub:using-platform-to-control-transitive-deps) that in turn [imports](https://docs.gradle.org/current/userguide/java_platform_plugin.html#sec:java_platform_bom_import) the [BOMs for Spring and Spring Boot](https://docs.spring.io/spring-boot/docs/2.7.8/gradle-plugin/reference/htmlsingle/#managing-dependencies.gradle-bom-support). The project is still on Java 11, so for the time being I need to [constrain](https://docs.gradle.org/current/userguide/dependency_constraints.html#sec:adding-constraints-transitive-deps) and [align](https://docs.gradle.org/current/userguide/dependency_version_alignment.html#sec:align_bom_platform) my Spring and Spring Boot versions to be less than 6.x and 3.x respectively, as those require Java 17.

This may be user error, but I'm unable to get the Gradle Versions Plugin to honor my platform constraints. Reproducer here:

https://github.com/ianbrandt/gradle-versions-plugin-platform-constraints/tree/f5b4360c2c5725cce61fa1ea0ed7d307ef3e8ed0

Constraint checking enabled:

https://github.com/ianbrandt/gradle-versions-plugin-platform-constraints/blob/f5b4360c2c5725cce61fa1ea0ed7d307ef3e8ed0/build.gradle.kts#L26

My platform, with Spring BOM import, and version alignment via a `ComponentMetadataRule` (I'm not sure if the rule can be declared in the platform, or if it should be declared in my convention plugin, but I tried it both ways and it didn't seem to make a difference):

https://github.com/ianbrandt/gradle-versions-plugin-platform-constraints/blob/f5b4360c2c5725cce61fa1ea0ed7d307ef3e8ed0/platforms/app-platform/build.gradle.kts

The version range constraints specified with [strict version shorthand](https://docs.gradle.org/current/userguide/single_versions.html#simple_version_declaration_semantics):

https://github.com/ianbrandt/gradle-versions-plugin-platform-constraints/blob/f5b4360c2c5725cce61fa1ea0ed7d307ef3e8ed0/gradle/libs.versions.toml#L13-L14

[...]

PR #1002, opened 2026-07-25
Fix: only override an inherited target jvm version in the query copy

### LINKED ISSUE #746: Running dependencyUpdates in projects without jvm-ecosystem can fail to resolve updated versions
# Steps to Reproduce

1. Checkout https://github.com/ben-manes/caffeine 
2. Run `dependencyUpdates`
3. See output:

```
> Task :dependencyUpdates

...

...

 - com.beust:jcommander
     1.82
```

Running with info shows:
```
 - com.beust:jcommander
     1.82
The exception that is the cause of unresolved state: org.gradle.internal.resolve.ModuleVersionResolveException: Could not resolve com.beust:jcommander:+.
Required by:
    project :
Caused by: org.gradle.internal.component.NoMatchingConfigurationSelectionException: No matching variant of com.beust:jcommander:1.82 was found. The consumer was configured to find attribute 'org.gradle.jvm.version' with value '2147483647' but:
  - Variant 'apiElements' capability com.beust:jcommander:1.82:
      - Incompatible because this component declares attribute 'org.gradle.jvm.version' with value '8' and the consumer needed attribute 'org.gradle.jvm.version' with value '2147483647'
  - Variant 'runtimeElements' capability com.beust:jcommander:1.82:
      - Incompatible because this component declares attribute 'org.gradle.jvm.version' with value '8' and the consumer needed attribute 'org.gradle.jvm.version' with value '2147483647'
```

# Resolution

In general, any project which performs dependency-resolution on JVM artifacts must apply the `jvm-ecosystem` plugin. The versions plugin sets the `TARGET_JVM_VERSION_ATTRIBUTE` attribute and therefore musta apply this plugin. 

To see the resolution in action, add `apply plugin: 'jvm-ecosystem'` to the root `build.gradle` file in caffeine. 

Context: https://github.com/ben-manes/gradle-versions-plugin/issues/727#issuecomment-1467157595

ianbrandt:
Fixes #746.

The query copies unconditionally requested `org.gradle.jvm.version` to disable auto target jvm (#727), but Gradle registers the compatibility rule for that attribute only in `JvmEcosystemPlugin`. In a project that applies no jvm plugin the request is uninterpretable, so a jvm dependency fails variant selection and the report lists it under "Failed to determine the latest version". Overriding the attribute only where the copy inherited one keeps the behavior for jvm projects and stops injecting the request into projects that never asked for it. The regression test resolves a new fixture module published with Gradle metadata from a project with no jvm plugin applied.

### DIFF
--- a/gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/Aggregation.kt
+++ b/gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/Aggregation.kt
@@ -8,7 +8,6 @@ import org.gradle.api.artifacts.component.ProjectComponentIdentifier
 import org.gradle.api.attributes.Category
 import org.gradle.api.attributes.VerificationType
 import org.gradle.api.internal.StartParameterInternal
-import org.gradle.api.plugins.JvmEcosystemPlugin
 import org.gradle.api.provider.Provider
 import org.gradle.api.services.BuildService
 import org.gradle.api.services.BuildServiceParameters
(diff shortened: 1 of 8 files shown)


ben-manes (approved):
hopefully its benign to sub-projects that don't, like empty ones in a hierarchy or nodejs/native ones.

ianbrandt:
> hopefully its benign to sub-projects that don't, like empty ones in a hierarchy or nodejs/native ones.

Good catch, thanks! Switched to a guard.

Issue #1006, opened 2026-07-26
`dependencyUpdates` fails on a composite build with configure on demand


ianbrandt:
A composite build with `org.gradle.configureondemand=true` fails the task outright. #1005 fixed the composite failure in #1004 by moving the locking deactivation into `projectsEvaluated`, but under configure on demand the task graph is computed before that callback fires, so the configuration is already resolved when it runs. This is on master with #1005 merged.

```
* What went wrong:
Cannot mutate the resolution strategy of configuration
':aggregateDependencyUpdatesResults' after the configuration was resolved.
After a configuration has been observed, it should not be modified.
```

## Reproducer

`settings.gradle`:

```groovy
includeBuild 'child'
```

`build.gradle`:

```groovy
plugins {
  id 'java-library'
  id 'io.github.ben-manes.versions'
}

repositories { mavenCentral() }

dependencies {
  api 'com.google.inject:guice:2.0'
}
```

`gradle.properties`:

```properties
org.gradle.configureondemand=true
```

`child/settings.gradle` and `child/build.gradle` are both empty.

## Trigger

Both halves are needed, on Gradle 9.6.1:

| shape | result |
| --- | --- |
| composite + configure on demand | fails as above |
| composite, `--no-configure-on-demand` | passes |
| configure on demand, no `includeBuild` | passes |

The configuration cache is not involved; the reproducer never enables it. On Gradle 8.4 the same site trips the neighbouring guard instead ("after task dependencies have been resolved"), which is the wording #1004 reported.

Easy to hit without meaning to: an `includeBuild` for `build-logic` plus `org.gradle.configureondemand=true` is enough, and root-only application by a catalog alias is enough. Three of my own multi-project builds match it.

Issue #1004, opened 2026-07-26
`dependencyUpdates` fails on any composite build since 0.55.0


ianbrandt:
A build with an `includeBuild` in its settings fails the task outright. 0.54.0 reports normally on the same project; 0.55.0 does not, so this is a regression from the aggregation rewrite in #994.

## Reproducer

`settings.gradle`:

```groovy
includeBuild 'child'
```

`build.gradle`:

```groovy
plugins {
  id 'java-library'
  id 'com.github.ben-manes.versions' version '0.55.0'
}

repositories { mavenCentral() }

dependencies {
  api 'com.google.inject:guice:2.0'
}
```

`child/settings.gradle` and `child/build.gradle` are both empty.

Actual (plugin 0.55.0, Gradle 8.4):

```
* What went wrong:
Execution failed for task ':dependencyUpdates'.
> Cannot change resolution strategy of dependency configuration
  ':aggregateDependencyUpdatesResults' after task dependencies have been resolved
```

Expected, and what 0.54.0 prints:

```
The following dependencies have later milestone versions:
 - com.google.inject:guice [2.0 -> 3.1]
```

## Cause

`registerAggregation` deactivates dependency locking on `aggregateDependencyUpdatesResults` from inside the provider that supplies the task's partial results (Aggregation.kt:146, added for #781). A composite build computes the task graph across the included builds, which resolves that configuration's dependencies and freezes its resolution strategy before the provider runs.

PR #1008, opened 2026-07-26
feat: report the plugin versions declared in the settings script

### LINKED ISSUE #367: Versions of plugins from settings.gradle are not checked

### LINKED ISSUE #533: Merged Report Missing
Hello,
I have a multi-module project that uses the `buildSrc` with convention files.  I have no `build.gradle.kts` in the root of the project for example.  In this setup, the dependency plugin generates a report into each subproject (`<subproject>/build/reports/dependencyUpdates`) however there is no merged result generated.  Probably because there is no root project in this style.

Gradle 7.0.2 with the Kotlin DSL
Version 0.39.0 of this plugin

### Project layout
root
  - gradle.properties
  - settings.gradle.kts
  - buildSrc
    - build.gradle.kts
    - src/main/kotlin/convention.kts files
  - project-A
    - build.gradle.kts
  - project-B
    - build.gradle.kts
  - project-C
    - build.gradle.kts

This is not a blocker since I do get individual reports.

Cheers,
John

### LINKED ISSUE #720: checking for gradle enterprise plugin version in settings.gradle.kts file
hello, when I execute the task  :dependencyUpdates , it does not check the gradle enterprise version in the settings.gradle.kts file

settings.gradle.kts:

```kotlin
plugins {
  id("com.gradle.enterprise").version("3.12.2")
}

gradleEnterprise {
    buildScan {
        termsOfServiceUrl = "https://gradle.com/terms-of-service"
        termsOfServiceAgree = "yes"
    }
}
```

ianbrandt:
Fixes #367, and closes #720 as a duplicate of it. Also fixes #533.

I tried a spike of a settings plugin for #367. Built on top of #994/#995, it worked out better than I expected.

This PR adds a new `io.github.ben-manes.versions.settings` plugin. Applied once in the settings script, it covers every project of the build. The root receives the `dependencyUpdates` task, and every other project contributes to it through the contributor plugin from #995. Invoking `dependencyUpdates` by name still runs one task and writes one merged report. No root build script is required (#533), and isolated projects needs no per-project convention wiring. It also reports the versions of the plugins the settings script itself declares (#367), which needed no new resolution machinery.

One note on build classpath pollution: a settings plugin puts itself and its dependencies (OkHttp, Moshi, and the Kotlin standard library) on the classpath of every build script in the build. To minimize the impact, only the Gradle API and the standard library execute in settings scope: settings evaluation records the settings classpath unresolved and hands it to the root project's task through the plugin's existing build service, so OkHttp and Moshi never load there. I considered resolving those dependencies at task execution instead, the way the code-quality plugins keep tool classpaths off the build script, but that would restructure the whole plugin and make every build's repositories responsible for serving them. So the dependencies ride the settings classpath as unloaded bytes, and the residual risk is a version conflict with another settings plugin wanting different versions. If that ever materializes, the durable fix is shading them into the jar under relocated packages, which is how the major settings plugins avoid the problem.

I've tested this out in a few projects I have, one of which is a somewhat involved composite build of a Gradle plugin. This uncovered #1004 and #1006, both now fixed in master.

I restructured the [README](https://github.com/ianbrandt/gradle-versions-plugin/blob/1ef973282a4426bf004c775c777a89631d428baa/README.md) to recommend the settings plugin as the standard approach. Root and per-project application moves to an "Other ways to apply the plugin" section. I tested those approaches against these changes via my three test projects, and all continue to work as expected. I also took a stab at an overall update of the README to make it current and hopefully more helpful to Gradle users of all skill levels.

The composite-build section now also shows the lifecycle-task fan-out for reporting on every included build in one invocation, since a report cannot span builds.

I added a "Migrating from prior versions" section that covers the user-facing changes from v0.54.0 and earlier, to v0.55.0 (still pending portal publication per #997), to what this PR would entail if merged and released as a new version.

Speaking of releasing, would you ever want to cut a v1.0.0, and possibly also adopt [SemVer](https://semver.org)? I don't have a strong preference myself, so that's not meant as a request or suggestion; I just wanted to ask the question before barrelling ahead with a v0.56.0.

If we do merge and release this, I believe the new plugin ID will trigger a manual Gradle Plugin Portal review. I learned my lesson with v0.55.0, and would delay creating the GitHub release until the publication is complete.

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,8 +226,9 @@ plugins {
 </details>
 
 The root project receives the `dependencyUpdates` task and every other project contributes to its
-report, so no project applies a plugin itself and a root build script is not required. Because
-the settings plugin puts the plugin on every project's classpath, a
+report, so no project applies a plugin itself and a root build script is not required. It also
+reports the versions of the plugins that the settings script declares, which no project's
+buildscript carries. Because the settings plugin puts the plugin on every project's classpath, a
 project that applies `io.github.ben-manes.versions` as well must request it without a version.
 
 The report is aggregated from a task in each project, so it is compatible with parallel execution
--- a/README.md
+++ b/README.md
@@ -198,8 +198,37 @@ field explaining failures or missing information. The update check may be disabl
 #### Multi-project build
 
 In a multi-project build, running this task in the root project will generate a consolidated/merged
-report for dependency updates in all subprojects. Alternatively, you can run the task separately in
-each subproject to generate separate reports for each subproject.
+report for dependency updates in all subprojects. The recommended setup is to apply the settings
+plugin once in the settings script, which covers every project of the build:
+
+<details open>
+<summary>Kotlin</summary>
+
+```kotlin
+// settings.gradle.kts
+plugins {
+  id("io.github.ben-manes.versions.settings")
+}
+```
+
+</details>
+
+<details>
+<summary>Groovy</summary>
+
+```groovy
+// settings.gradle
+plugins {
+  id 'io.github.ben-manes.versions.settings'
+}
+```
+
+</details>
+
+The root project receives the `dependencyUpdates` task and every other project contributes to its
+report, so no project applies a plugin itself and a root build script is not required. Because
+the settings plugin puts the plugin on every project's classpath, a
+project that applies `io.github.ben-manes.versions` as well must request it without a version.
 
 The report is aggregated from a task in each project, so it is compatible with parallel execution
 and the configuration cache. Each project takes the settings that control resolution (`revision`,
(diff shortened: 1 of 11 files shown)


ben-manes (approved):
Can you try following up with a dependency updates check of the settings' dependencies (per prior ticket discussion)?

Technically we do use semver since 0.x signals that I had no clue what I was doing 🙂. It was a mixture of mostly works, no compatibility promises, best effort support, open for contributions, and no ego if people prefer something else. I suppose plugin development feels too fluid most of the time that patch/features/compatibility seem awkward since the underlying tool is changing a lot underneath us.

ianbrandt:
> Can you try following up with a dependency updates check of the settings' dependencies (per prior ticket discussion)?

I could try. One question: the settings buildscript's `classpath` dependencies are already reported alongside the declared plugins. That leaves plugins pinned in `pluginManagement { plugins { ... } }` but never applied. I left those out deliberately. Per @Vampire's https://github.com/ben-manes/gradle-versions-plugin/issues/367#issuecomment-1804247687, a dormant pin is a version constraint rather than a dependency, and once any script applies the plugin it reaches a real classpath and is reported. A test covers the exclusion. Would you want the dormant pins reported anyway? 

They aren't a configuration, so covering them would take your detached-configuration approach (https://github.com/ben-manes/gradle-versions-plugin/issues/367#issuecomment-604751717), resolved against `pluginManagement.repositories` with its resolution strategy.

If so, I'll attempt a follow-up PR.

> Technically we do use semver since 0.x signals that I had no clue what I was doing 🙂. It was a mixture of mostly works, no compatibility promises, best effort support, open for contributions, and no ego if people prefer something else. I suppose plugin development feels too fluid most of the time that patch/features/compatibility seem awkward since the underlying tool is changing a lot underneath us.

Ha, makes sense, and I do see how SerVer could be potentially tricky to track for a plugin vs. a library. So, just to confirm, v0.56.0 for this if we release it then?

ben-manes:
I think `pluginManagement.plugins` can be ignored. I was more interested in settings plugins that cannot be sourced from the version catalog yet ([example](https://github.com/ben-manes/caffeine/blob/0f9cf9d1b6cce6de326443820b820b015ffb0ddb/settings.gradle.kts#L15-L19)).

v0.56.0 sgtm!

ianbrandt:
> I think `pluginManagement.plugins` can be ignored. I was more interested in settings plugins that cannot be sourced from the version catalog yet ([example](https://github.com/ben-manes/caffeine/blob/0f9cf9d1b6cce6de326443820b820b015ffb0ddb/settings.gradle.kts#L15-L19)).

Those are covered. The settings script's `plugins` block lands on its buildscript classpath, which the report now includes, so all three would show with their declared versions. Two of them were in the builds I tested against—`com.gradle.develocity` and the foojay resolver, declared exactly that way—and both were reported.

> v0.56.0 sgtm!

👍

PR #1026, opened 2026-07-29
Do not publish a variant to a project that publishes by fallback (fixes #1022)

### LINKED ISSUE #1022: Regression [0.54.0 -> 0.55.0] Could not resolve project
In our Projekt (https://github.com/Governikus/AusweisApp-SDK-Wrapper) everything works fine with 0.54.0.

* As example `./gradlew assembleDebug` works great in the `android` folder.
* It also works, when we place [ausweisapp.aar](https://repo1.maven.org/maven2/com/governikus/ausweisapp/2.5.4/ausweisapp-2.5.4.aar) in `android/ausweisapp`. `sdkwrapper` and `tester` will use the local aar in that case instaed of `com.governikus:ausweisapp`.

When we update to 0.55.0 or 0.56.0 in settings.gradle, it still works great, when we do NOT provide a local aar. With a local aar we run into the following error:

```
./gradlew assembleDebug

> Configure project :
The com.github.ben-manes.versions plugin id is deprecated; apply io.github.ben-manes.versions instead.

> Configure project :sdkwrapper
⚠   ⚠   ⚠   ⚠   WARNING  ⚠   ⚠   ⚠   ⚠
   Using local SDK might be outdated  
⚠  Delete local SDK in lib folder to ⚠
   use latest version!                
⚠   ⚠   ⚠   ⚠   WARNING  ⚠   ⚠   ⚠   ⚠

> Configure project :tester
⚠   ⚠   ⚠   ⚠   WARNING  ⚠   ⚠   ⚠   ⚠
   Using local SDK might be outdated  
⚠  Delete local SDK in lib folder to ⚠
   use latest version!                
⚠   ⚠   ⚠   ⚠   WARNING  ⚠   ⚠   ⚠   ⚠

[Incubating] Problems report is available at: file:///home/user/Dokumente/AusweisApp-SDKWrapper/android/build/reports/problems/problems-report.html

FAILURE: Build failed with an exception.

* What went wrong:
Could not determine the dependencies of task ':sdkwrapper:checkDebugAarMetadata'.
> Could not resolve all dependencies for configuration ':sdkwrapper:debugRuntimeClasspath'.
   > Could not resolve project :ausweisapp.
     Required by:
         project :sdkwrapper
      > No matching variant of project :ausweisapp was found. The consumer was configured to find a library for use during runtime, preferably optimized for Android, as well as attribute 'com.android.build.api.attributes.AgpVersionAttr' with value '8.13.2', attribute 'com.android.build.api.attributes.BuildTypeAttr' with value 'debug', attribute 'org.jetbrains.kotlin.platform.type' with value 'androidJvm' but:
          - Variant 'dependencyUpdatesElements':
              - Incompatible because this component declares a component of category 'verification' and the consumer needed a library
              - Other compatible attributes:
                  - Doesn't say anything about com.android.build.api.attributes.AgpVersionAttr (required '8.13.2')
                  - Doesn't say anything about com.android.build.api.attributes.BuildTypeAttr (required 'debug')
                  - Doesn't say anything about its target Java environment (preferred optimized for Android)
                  - Doesn't say anything about its usage (required runtime)
                  - Doesn't say anything about org.jetbrains.kotlin.platform.type (required 'androidJvm')

* Try:
> No matching variant errors are explained in more detail at https://docs.gradle.org/8.14.5/userguide/variant_model.html#sub:variant-no-match.
> Review the variant matching algorithm at https://docs.gradle.org/8.14.5/userguide/variant_attributes.html#sec:abm_algorithm.
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.
> Get more help at https://help.gradle.org.

Deprecated Gradle features were used in this build, making it incompatible with Gradle 9.0.

You can use '--warning-mode all' to show the individual deprecation warnings and determine if they come from your own scripts or plugins.

For more on this, please refer to https://docs.gradle.org/8.14.5/userguide/command_line_interface.html#sec:command_line_warnings in the Gradle documentation.

BUILD FAILED in 5s
```

ben-manes:
Fixes #1022, a regression introduced in 0.55.0 by the aggregation feature (#948).

## The bug

A project that declares no variant of its own — the flat-file pattern below, which the reporter's build uses to swap in a local aar — is resolved by falling back to its `default` configuration whatever the consumer asks for:

```groovy
configurations.maybeCreate('default')
artifacts.add('default', file('ausweisapp.aar'))
```

Gradle drops that fallback as soon as the project declares any variant. Since 0.55.0 the plugin adds a `dependencyUpdatesElements` variant to every project, which flipped such projects into variant-aware matching and left their consumers with no match:

```
> Could not resolve project :ausweisapp.
   > No matching variant of project :ausweisapp was found. ...
      - Variant 'dependencyUpdatesElements':
          - Incompatible because this component declares a component of category 'verification'
            and the consumer needed a library
```

## The fix

Two parts, each independently necessary:

1. The variant is published once the project is configured, and only when the project does not publish through its `default` configuration. The predicate reads that configuration's own artifacts rather than the absence of other variants, so that it is not sensitive to a plugin that declares its variants from its own `afterEvaluate` — verified against AGP, whose library and application projects both keep publishing the variant.
2. The aggregate names the producer's configuration instead of matching it by attributes, so a project without one is skipped rather than resolved through its `default` configuration, whose artifacts would then be read as a partial result.

A project that is skipped is still aggregated: its producer's output is wired as a task output.

## Verification

Reproduced against the reporter's build ([Governikus/AusweisApp-SDK-Wrapper](https://github.com/Governikus/AusweisApp-SDK-Wrapper)) with `ausweisapp-2.5.4.aar` in `android/ausweisapp`: `./gradlew assembleDebug` fails on 0.56.0 with the error above and succeeds with this change. Only `:ausweisapp` skips the variant there; `:sdkwrapper`, `:tester`, and the root project all publish it.

`DefaultConfigurationSpec` covers both halves — each case fails without its corresponding part of the fix, the first with the reported `No matching variant`, the second with `dependencyUpdates` reading the local jar as a partial result.

### DIFF
--- a/gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/Aggregation.kt
+++ b/gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/Aggregation.kt
@@ -4,6 +4,7 @@ import com.github.benmanes.gradle.versions.updates.resolutionstrategy.Resolution
 import org.gradle.api.Action
 import org.gradle.api.Project
 import org.gradle.api.artifacts.Configuration
+import org.gradle.api.artifacts.Dependency
 import org.gradle.api.artifacts.component.ProjectComponentIdentifier
 import org.gradle.api.attributes.Category
 import org.gradle.api.attributes.VerificationType
@@ -170,10 +171,15 @@ internal fun registerAggregation(
 
   // Declared for every project so that computing the aggregate's dependencies configures each of
   // them, which configure on demand skips when the task is invoked by its path rather than by name.
+  // The producer's configuration is named rather than matched by its attributes, so that a project
+  // which publishes no variant is skipped instead of falling back to its `default` configuration,
+  // whose artifacts are the project's own and not a partial result to read the report from.
   for (aggregated in project.allprojects) {
     project.dependencies.add(
       AGGREGATION_CONFIGURATION,
-      project.dependencies.project(mapOf("path" to aggregated.path)),
+      project.dependencies.project(
+        mapOf("path" to aggregated.path, "configuration" to ELEMENTS_CONFIGURATION),
+      ),
     )
   }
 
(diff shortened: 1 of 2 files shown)


ianbrandt (approved):
Thanks for fixing my regression!

PR #1035, opened 2026-07-31
Name the projects behind a coordinate that resolves to more than one version

ianbrandt:
Addresses #1032.

The merged report drops which project declared which version, so one coordinate can print at several versions with nothing telling them apart. Kafka's `streams:upgrade-system-tests-*` projects each declare a different old `kafka-streams` on purpose, and today they render like this:

```
 - org.apache.kafka:kafka-streams [0.11.0.3 -> 4.3.1]
     https://kafka.apache.org
 - org.apache.kafka:kafka-streams [1.0.2 -> 4.3.1]
     https://kafka.apache.org
 - org.apache.kafka:kafka-streams [1.1.1 -> 4.3.1]
     https://kafka.apache.org
```

This PR names the projects behind each version:

```
 - org.apache.kafka:kafka-streams [0.11.0.3 -> 4.3.1]
     https://kafka.apache.org
     declared in :streams:upgrade-system-tests-0110
 - org.apache.kafka:kafka-streams [1.0.2 -> 4.3.1]
     https://kafka.apache.org
     declared in :streams:upgrade-system-tests-10
 - org.apache.kafka:kafka-streams [1.1.1 -> 4.3.1]
     https://kafka.apache.org
     declared in :streams:upgrade-system-tests-11
```

The jacoco split in #1032 reads the same way, with the root applying `jacoco` and no `toolVersion` while the subprojects set 0.8.15. Kafka declares it in 65 projects, so this is also where the list gets long:

```
 - org.jacoco:org.jacoco.agent:0.8.15
     declared in :clients, :clients:clients-integration-tests, :connect, :connect:api, :connect:basic-auth-extension and 60 others
 - org.jacoco:org.jacoco.agent [0.8.14 -> 0.8.15]
     http://jacoco.org
     declared in root project
```

The plain text and HTML reports name the first five projects and count the rest; the JSON and XML reports carry the complete list, so nothing is lost for a tool reading the report. Uncapped, that first line runs about 900 characters and becomes a monstrous cell in the HTML table:

```
 - org.jacoco:org.jacoco.agent:0.8.15
[...]

Issue #1032, opened 2026-07-30
[Follow up / test for #1028] Checking does jacoco version resolution works as expected with plugin version 0.57.0


dejan2609:
**Note:** #1028 follow up / test.

Motivation/rationale (FYI @ianbrandt and @ben-manes): I just want to be sure that this is a desired behaviour.

**Environment:** 
- plugin version 0.57.0
- Gradle version: 9.4.1
- JDK: 25

**POC:** Apache Kafka trunk + few version changes: https://github.com/dejan2609/kafka/tree/testing-jacoco-version-resolution-with-versions-plugin-0.57.0

```
 git log -3 --oneline 
442120e47f4 (HEAD -> testing-jacoco-version-resolution-with-versions-plugin-0.57.0) commenting out both if statements that are wrapping jacoco configuration (in order to mimic spotbugs configuration)
4518adc7c0f version bumps: `io.github.ben-manes.versions` (0.56.0 -->> 0.57.0) and `jacoco` (0.8.14 -->> 0.8.15)
2cc6ef65e5f (upstream/trunk, upstream/HEAD, origin/trunk, origin/HEAD, trunk) KAFKA-20736: Close empty share sessions safely (#22888)
```

I'll just add few more things that probably should have been added to the previous issue description (chances are that you probably managed to decipher this on your own, but I just want to come up clean):
- both spotbugs and jacoco versions are defined via `gradle/dependencies.gradle`:
```
versions += [
  jacoco: "0.8.14",
  spotbugs: "4.9.8",
]
```
- `build.gradle` related plugin configuration goes like this:
```
ext {
  userEnableTestCoverage = project.hasProperty("enableTestCoverage") ? enableTestCoverage : false
}

subprojects {

    apply plugin: "com.github.spotbugs"
    spotbugs {
        toolVersion = versions.spotbugs
    }

    // Ignore core since its a scala project
    if (it.path != ':core') {
        if (userEnableTestCoverage) {
            apply plugin: "jacoco"

            jacoco {
                toolVersion = versions.jacoco
            }
        }
    } 
}  
```
- idea is to be able to generate test coverage for all non-Scala subprojects like this (see `README.md`):
  - coverage for an entire project: `./gradlew reportCoverage -PenableTestCoverage=true -Dorg.gradle.parallel=false` 
  - coverage for a single module `./gradlew clients:reportCoverage -PenableTestCoverage=true -Dorg.gradle.parallel=false`
  
  
### Test cases:
[...]

Issue #1028, opened 2026-07-29
[v0.56.0 / Gradle 9.4.1 / large project] Jacoco current version defaults to a Gradle built-in version


dejan2609:
**Issue description:** 
 - Jacoco: no matter what version is declared via `jacoco.toolVersion` current version defaults to 0.8.14 :x:
 - SpotBugs: everything works fine (i.e. `spotbugs.toolVersion` is honored) ✅ 

💡 **Hint**: Jacoco is a first-level citizen in Gradle (while SpotBugs is not).
 
**POC**: Apache Kafka trunk + few version changes for Jacoco and SpotBugs: https://github.com/dejan2609/kafka/tree/jacoco-current-version-resolution-issue 
```
git log -3 --oneline 
21c6d90cb94 (HEAD -> jacoco-current-version-resolution-issue) `jacoco.toolVersion` = 0.8.15, `spotbugs.toolVersion` = 4.10.3
c7ee2a40169 `jacoco.toolVersion` = 0.8.13, `spotbugs.toolVersion` = 4.9.7
209a1b6e39a (upstream/trunk, upstream/HEAD, origin/trunk, origin/HEAD, trunk) KAFKA-20328 : [1/N] headers-aware api test coverage (#22196)
```
  
### Test cases for `./gradlew dU | grep -E "org.jacoco:org.jacoco|com.github.spotbugs:spotbugs"`
1. Kafka trunk (commit `209a1b6e39a`) `jacoco.toolVersion` = 0.8.14, `spotbugs.toolVersion` = 4.9.8
```
 - com.github.spotbugs:spotbugs [4.9.8 -> 4.10.3]
 - org.jacoco:org.jacoco.agent [0.8.14 -> 0.8.15]
 - org.jacoco:org.jacoco.ant [0.8.14 -> 0.8.15]
```
2. commit `c7ee2a40169` - `jacoco.toolVersion` = 0.8.13, `spotbugs.toolVersion` = 4.9.7
```
 - com.github.spotbugs:spotbugs [4.9.7 -> 4.10.3]
 - org.jacoco:org.jacoco.agent [0.8.14 -> 0.8.15]
 - org.jacoco:org.jacoco.ant [0.8.14 -> 0.8.15]
```
3. commit `21c6d90cb94` - `jacoco.toolVersion` = 0.8.15, `spotbugs.toolVersion` = 4.10.3
```
 - com.github.spotbugs:spotbugs:4.10.3
[...]

Issue #844, opened 2024-03-05
Report has differences when nothing changes


dalewking:
Was trying to upgrade some dependencies and wanted to make sure that I was not getting unexpected transitive dependency updates. What i wanted to do was:

- run dependencyUpdates
- save the report
- upgrade a dependency
- run dependencyUpdates
- compare the previous report to see what changed

The problem is there were differences in the report even with no changes to dependencies.

It appears for the most of the changes it it that it is somewhat randomly choosing between the old and new version for the project URL

For example, here is part of the JSON report:

```
   {
    "group": "androidx.compose.ui",
    "name": "ui",
    "version": "1.4.3",
    "projectUrl": "https://developer.android.com/jetpack/androidx/releases/compose-ui#1.6.2",
    "userReason": null,
    "available": {
     "release": null,
     "milestone": "1.6.2",
     "integration": null
    }
   },
```

compared to this entry with no changes to dependencies:

```
   {
    "group": "androidx.compose.ui",
    "name": "ui",
    "version": "1.4.3",
    "projectUrl": "https://developer.android.com/jetpack/androidx/releases/compose-ui#1.4.3",
    "userReason": null,
    "available": {
     "release": null,
     "milestone": "1.6.2",
     "integration": null
    }
   },
```

I also see differences in reason:
```
   {
    "group": "androidx.test",
    "name": "rules",
    "version": "1.6.0-alpha01",
    "projectUrl": "1.5.0",
    "userReason": null,
    "reason": "org.gradle.internal.resolve.ModuleVersionResolveException: Could not resolve androidx.test:rules:{strictly 1.6.0-alpha01}.\nRequired by:....
```

vs.

```
   {
    "group": "androidx.test",
    "name": "rules",
    "version": "1.6.0-alpha01",
    "projectUrl": "1.5.0",
    "userReason": null,
    "reason": "org.gradle.internal.resolve.ModuleVersionResolveException: Could not resolve androidx.test:rules:+.\nRequired by:
```

ben-manes:
Since we defer to Gradle to resolve the dependency, maybe it is non-deterministic due to caching and repository ordering? I usually add `--refresh-dependencies` so I wonder if that would make it more consistent for you, as it forces it to bypass the cache?

dalewking:
I don't think it is gradle, because it is not a question of versions (at least for the first part). It is getting the correct versions for current and milestone versions, it is the projectUrl

ben-manes:
[...]

PR #1039, opened 2026-08-01
Prepare the 0.58.0 release

ianbrandt:
Bumps the version for the v0.58.0 release. Let me know if there's anything more you want included in this release.

Draft release notes below.

---

* Honor the build's log level in the reporters: the console summary prints at the lifecycle level, so `--quiet` suppresses it; a skipped report file is logged at the info level; the "use --info for details" hint is dropped when the build already runs at that level; and an unresolved dependency no longer reports a version string as its project url (#278, #388, #726, #844, #1033)
* Report a module published only as a snapshot as up to date, rather than as a failure to determine its latest version (#475, #1034)
* Skip the buildscript classpath of a project that declares no repository, which previously reported each of that project's buildscript dependencies as a failure to determine its latest version (#756, #1038)
* Name the projects that declared each version when one coordinate resolves to more than one, so the merged report of a multi-project build no longer prints the same dependency several times with nothing telling the entries apart (#1032, #1035)
* Mark a dependency that only a plugin contributed and name the configuration it came from, so a `jacoco`, `checkstyle` or `pmd` tool version the build never sets no longer reads as a resolution bug (#1028, #1032, #1036)

> [!NOTE]
>
> The console summary now honors the log level, so a script running the task under `--quiet` no longer sees it. The report file is still written. Report entries may also carry an indented attribution line naming the projects that declared a divergent version or the configuration a plugin contributed the dependency into, which a tool parsing the plain text report line by line has to skip. See [Migrating from prior versions](https://github.com/ben-manes/gradle-versions-plugin#v0570).

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ Plugin](https://www.mojohaus.org/versions-maven-plugin).
 - [Samples](#samples)
 - [Compatibility](#compatibility)
 - [Migrating from prior versions](#migrating-from-prior-versions)
+  - [v0.57.0](#v0570)
   - [v0.56.0](#v0560)
   - [v0.55.0](#v0550)
   - [v0.54.0 and earlier](#v0540-and-earlier)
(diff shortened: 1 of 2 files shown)


ben-manes (approved):

ben-manes:
```console
Caused by: java.lang.NoSuchMethodError: 'void com.github.benmanes.gradle.versions.reporter.PlainTextReporter.<init>(java.lang.String, java.lang.String, java.lang.String, boolean, int, kotlin.jvm.internal.DefaultConstructorMarker)'
        at Versions_caffeine_gradle$1$3.execute(versions.caffeine.gradle.kts:37)
```

I tried updating caffeine and got this?

ianbrandt:
Uh oh. Looking into it...

ianbrandt:
Could `settings.gradle.kts` still be on 0.57.0 while `libs.versions.toml` is on 0.58.0? The settings classpath is the parent, so `versions.caffeine.gradle.kts` would compile against 0.58.0 and link against 0.57.0 at runtime.

ben-manes:
oh whoops! sorry, I was only half paying attention and working on other stuff. That was it, thx!

PR #1041, opened 2026-08-01
Write the partial results under the project that aggregates them (fixes #1040)

### LINKED ISSUE #1040: A project with no build script still gets a build/dependencyUpdates/partial.json
### What happens

Since 0.55.0 every project in the build gets a `partialDependencyUpdates` producer, including the
intermediate container projects that Gradle creates implicitly for a nested `include`. Such a project
has no build script and can therefore declare nothing, yet it still gets a `build/` directory holding
an empty partial result:

```json
{"formatVersion":1,"projectPath":":apps","statuses":[],"buildscriptStatuses":[]}
```

### Reproduction

`settings.gradle.kts`:

```kotlin
include(":apps:app-a")
include(":apps:app-b")
```

`:apps` exists as a project but has no `build.gradle.kts` of its own. Then:

```
./gradlew dependencyUpdates
```

leaves behind

```
apps/build/dependencyUpdates/partial.json
```

In our build that is six new directories — `apps/`, `ui/`, `infrastructure/`, `presentation/`,
`testing/`, `baselineprofile/`. Each shows up as untracked in git, since the per-module `.gitignore`
files exist only for the real modules; the container directories had never produced any output before.

### Why this looks unintended rather than deliberate

`Aggregation.kt` already filters exactly these projects out when it builds the set for the
completeness warning:

```kotlin
val aggregatedPaths =
  project.allprojects.filter { it.buildFile.exists() }.map { it.path }.toSet()
```

The producer registration a few lines below does not apply the same filter:

```kotlin
project.allprojects { aggregated ->
  if (claims(aggregated)) {
    val partial = registerProducer(aggregated, service)
```

The aggregated report is byte-for-byte identical either way, because the partial results of these
projects are empty.

### Environment

- gradle-versions-plugin 0.57.0, plugin id `io.github.ben-manes.versions`, applied to the root
  project through a precompiled convention plugin
- Gradle 9.6.1, configuration cache enabled
- JDK 21, Windows

Not a blocker — one `.gitignore` line covers it. Filing it because the two code paths disagree about
which projects are worth visiting.

ianbrandt:
Fixes #1040.

This PR writes the partial results under the project that aggregates them instead of each project's own build directory. For the issue's reproducer, `./gradlew dependencyUpdates` used to leave:

```
apps/app-a/build/dependencyUpdates/partial.json
apps/app-b/build/dependencyUpdates/partial.json
apps/build/dependencyUpdates/partial.json
build/dependencyUpdates/partial.json
```

and now leaves only:

```
build/dependencyUpdates/partials/apps-35f1dac.json
build/dependencyUpdates/partials/apps-app-a-c809bc67.json
build/dependencyUpdates/partials/apps-app-b-c809bc68.json
build/dependencyUpdates/partials/root-3a.json
```

I didn't filter the producers to the projects with a build script, because a build can declare dependencies on a container project (`project(':apps') { dependencies { ... } }`) and skipping its producer would silently drop them from the report. A `--clean-legacy-partials` option removes what earlier releases left behind, which `clean` cannot reach in a project that applies no plugin of its own.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ Plugin](https://www.mojohaus.org/versions-maven-plugin).
 - [Samples](#samples)
 - [Compatibility](#compatibility)
 - [Migrating from prior versions](#migrating-from-prior-versions)
+  - [v0.58.0](#v0580)
   - [v0.57.0](#v0570)
   - [v0.56.0](#v0560)
   - [v0.55.0](#v0550)
@@ -1511,6 +1512,17 @@ projects (see [Isolated projects](#isolated-projects)) are supported.
 Start at the subsection for the version your build is on and work upward: each
 subsection migrates to the version covered by the subsection above it, and the
 topmost migrates to the current release.
+
+### v0.58.0
+
+v0.59.0 collects the partial result of every project under the project that
+aggregates them, so a project that exists only to hold a nested `include` no
+longer gains a `build` directory of its own:
+
+* Run `./gradlew dependencyUpdates --clean-legacy-partials` once to remove the
+  `build/dependencyUpdates/partial.json` that earlier releases wrote into each
+  project. `clean` removes it from a project that applies a plugin of its own,
+  but a project with no build script has no `clean` task to reach it.
 
 ### v0.57.0
 
--- a/README.md
+++ b/README.md
@@ -1159,6 +1159,35 @@ configuration cache the dependency metadata read during resolution is tracked as
 an input, so a newly published version invalidates the cached entry rather than
 serving a stale report.
 
+The merged report and the partial results it merges live in the aggregating
+project's build directory. `clean` removes them only if that project has a
+`clean` task, which the `base` plugin supplies. A root project that applies no
+other plugin can add `base` for that alone; the task itself does not need it.
+
+<details open>
+<summary>Kotlin</summary>
+
+"build.gradle.kts":
+```kotlin
+plugins {
+  base
+}
+```
+
+</details>
+
+<details>
+<summary>Groovy</summary>
+
+"build.gradle":
+```groovy
+plugins {
+  id 'base'
+}
+```
+
+</details>
+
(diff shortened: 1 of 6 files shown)


ianbrandt:
For prior art, I looked at how some other aggregating plugins handle their per-project intermediates:

- Gradle's [`jacoco-report-aggregation`](https://docs.gradle.org/current/userguide/jacoco_report_aggregation_plugin.html) and [`test-report-aggregation`](https://docs.gradle.org/current/userguide/test_report_aggregation_plugin.html) are applied to one dedicated project, and the aggregated projects are declared explicitly, e.g. `jacocoAggregation(project(":app"))`. The aggregation side creates no tasks or files of its own; it reads what each producer's own tasks already wrote to that project's build directory.
- [Kover](https://github.com/Kotlin/kotlinx-kover) merges opt-in per project too, via `kover(project(":core"))` dependencies. [kotlinx-kover#706](https://github.com/Kotlin/kotlinx-kover/issues/706) looks like an equivalent report to #1040 from a build that opted in via `allprojects`, and was closed as not planned.
- [Dokka](https://github.com/Kotlin/dokka)'s multi-module tasks exist only where the plugin is applied.
- The [dependency analysis plugin](https://github.com/autonomousapps/dependency-analysis-gradle-plugin) aggregates to the root but keeps its intermediates in each project.
- [SonarScanner for Gradle](https://docs.sonarsource.com/sonarqube-server/analyzing-source-code/scanners/sonarscanner-for-gradle/) is applied at the root only, though its per-project working directory writes into every covered project when configured broadly.

The common pattern is opt-in per project. Projects that don't apply producers don't get `build/` outputs. Applying this plugin to the root project has always covered every subproject, since before the v0.55.0 rework. The settings plugin retains that feature and also reports the plugins the settings script itself declares, which a per-project opt-in cannot see. Centralizing the producers' output to the project that aggregates them seemed the next best thing.

ben-manes (approved):

PR #1042, opened 2026-08-01
Add a contributing guide and verify the release version on deploy

ianbrandt:
I thought it might be helpful to have a `CONTRIBUTING.md` for capturing the release process and other development conventions. This is just a draft based on my guesses at the de facto conventions, so please feel free to mark it up with anything that seems off or should be changed.

I also hardened the deploy workflow to refuse publishing when `VERSION_NAME` does not match the ref, so a manual dispatch that forgets `--ref` fails instead of publishing `master`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,7 +7,7 @@ against `master`.
 
 - Every push to the pull request runs the
   [build workflow](.github/workflows/build.yml) on JDK 8, 11 and 17. A change
-  touching only `README.md` skips it.
+  touching only `README.md` or `CONTRIBUTING.md` skips it.
 - Reference the issue a change addresses in the commit summary and in the
   pull request body, so the merge closes it:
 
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,191 @@
...
+  is rebased where the commit message should land word for word.
+- Merge commits are not used.
+
+## Building and testing
+
+- `./gradlew build` compiles the plugin and runs its Spock specs, including
+  the functional tests that drive real Gradle builds through TestKit.
+- On Windows, `gradlew.bat` stands in for `./gradlew` throughout.
+- The plugin targets Java 8 bytecode, so a change that needs a newer API
+  needs a guard.
```
ben-manes: fwiw, this could be increased to match the minimum Gradle's supported JVM since the plugin runs in its process.

ben-manes (approved):

CONTRIBUTING.md
ianbrandt: I believe that would be Java 8, as our minimum supported [Gradle version of 8.4 still supports it](https://docs.gradle.org/8.4/userguide/compatibility.html). I updated the `CONTRIBUTING.md` to clarify that.

PR #1045, opened 2026-08-02
Align the migration section's alerts with the release notes convention

ianbrandt:
Aligned alert usage in the README migration section with the release notes convention, and scoped the custom `outputFormatter` constraint to the configuration cache, where it actually applies.


ianbrandt:
I went ahead and merged this without review because it's just documentation cleanup, and I was stacking some local v0.59.0 release prep on it. I'll happily incorporate any retroactive feedback in a follow-up.

PR #1064, opened 2026-08-06
Test and document Gradle 9.7.0 as the current release

ianbrandt:
9.7.0 shipped and no release candidate follows it, so the specs that pinned 9.7.0-rc-2 run against the release instead. The sample reports were regenerated rather than edited: the breadcrumb loses a segment instead of gaining a version, since the release candidate check now reports no version at all and never compares greater than the current release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,7 +111,7 @@ The following dependencies have later milestone versions:
      https://spring.io/projects/spring-boot
 
 Gradle release-candidate updates:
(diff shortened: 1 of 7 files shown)


ianbrandt:
It looks like [Actions are down](https://www.githubstatus.com/incidents/qcvjkzcs7j74). Holding on any PR merges until they're back and we get green builds.

PR #1085, opened 2026-08-29
Accept the task settings as command line options

ianbrandt:
This PR adds a command line option to every `dependencyUpdates` property with a plain value, so a single run can change what the report contains without an edit to the build script. Where more than one is set, the command line option is the one that applies, ahead of the system property of the same name and of what is configured in the build. There is no option for the properties configured with a predicate, since no command line can express the logic. An option applies only within the build it is invoked in, so in a report that merges an included build, that build's rows are resolved from that build's own configuration, as documented beside the option table in the README.

I'd suggest task options as the preferred execution-time configuration approach over system properties going forward. They appear in `gradle help --task dependencyUpdates`, a misspelled option fails the build where `-DcheckConstraint=true` is silently ignored, and each is scoped to the task rather than to the whole JVM. The five existing system properties are unchanged.

Both a build script assignment and a command line option are applied through `Property.set()`. The system property is read between them, and there is nowhere in a `Property` to put it, so each option is backed by a separate field rather than by a `Property`. The members these options are declared on are `internal`, so moving them to the Provider API later would not be visible to a build.

If at some point in the future we find the command line options to be a suitable replacement, we could consider deprecating the system properties for removal by printing a warning when they're used that points to migration instructions. Removing them outright in this release is the alternative, and I could do that here instead. Then we could move forward with the [Provider API](https://github.com/gradle/build-tool-roadmap/issues/28) for the options now. I kept the system properties to be conservative and not force a significant migration on existing builds and build logic.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -61,6 +61,30 @@ How a pull request lands:
 - Beyond that, match the file you are editing: Kotlin sources carry no
   license header, comments are rare and explain only what the code cannot,
   and a README snippet is added in both Kotlin and Groovy.
+- Add a command line option, with `@Option`, to every task property a build
+  author sets to configure the report or where it is written, so that a single
+  run can change it without an edit to the build script. A property configured
+  with a predicate has no option, since no command line can express the logic.
+  The system properties named after `revision`, `gradleReleaseChannel`,
+  `outputFormatter`, `outputDir` and `reportfileName` predate the options and
+  remain for compatibility; where more than one is set, the option is the one
+  that applies, and no system property is added for a new property.
+- Gradle applies a command line option by calling a setter on the task. Have
+  that setter assign a separate field, not the field assigned from a build
+  script, and read them in order: the command line value first, then the system
+  property if there is one, then the configured value. Assigning the same field
+  fails three ways. A system property is read before the configured value, so
+  the value from `-Drevision` would be used instead of the one from
+  `--revision`. The configured value is read from the nearest project up the
+  hierarchy where one is set, so an option given on the root task would not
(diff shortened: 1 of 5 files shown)


ben-manes (approved):
tbh I'm not sure if the CLI options are very useful nowadays. Ivy's `revision` was the only way to manage resolution in the earliest days when Gradle was a light wrapper above it. I think this is fine and nice when exploring options like constraints, just wouldn't expect many usages that weren't from misunderstandings.

ianbrandt:
> tbh I'm not sure if the CLI options are very useful nowadays. Ivy's `revision` was the only way to manage resolution in the earliest days when Gradle was a light wrapper above it. I think this is fine and nice when exploring options like constraints, just wouldn't expect many usages that weren't from misunderstandings.

Agreed, and I actually thought about not including `--revision`; `--gradle-versions-api-base-url` too, as it probably has a minority use case (testing a proxy or mirror, maybe 🤷‍♂️). At the same time, leaving them out seemed like a bit of an asymmetry, so I just went ahead and included them all. Happy to remove one or both if you prefer.

Spoiler alert, this PR fell out of another one I'm looking into where I may have a good use case for a CLI option. Stay tuned. 😁 Once I got it in my head that task options seemed a better approach than system properties (for the reasons mentioned above), this PR seemed worth doing either way.

It looks like the build failed on the older JDKs. Looking into that now...

ianbrandt:
Build fixed, and I put some more effort into the README to help combat misunderstandings. Let me know if you'd like to see some changes, or if you think this is good as is.

ben-manes:
lgtm

Issue #798, opened 2023-07-25
Plugin not show the latest stable version when there is alpha version


westito:
I have declared `com.squareup.okhttp3:okhttp:4.9.3` in my build.gradle file. When no version filter added, it prints there is a new alpha (5.0.0-alpha.11) version. So I added the version rejecter described in readme.md. As the result, it is not reporting outdated, but also not reports the newer `com.squareup.okhttp3:okhttp:4.11.0` version. It says 
```
The following dependencies are using the latest milestone version:
- com.squareup.okhttp3:okhttp:4.9.2
```
But it isn't true! I add a debug printer in the `rejectVersionIf` script and it is returns 
```
isNonStable: true 5.0.0-alpha.1
isNonStable: false 4.11.0
```
So it seems this script working, somehow the plugin rejects stable version too if alpha rejected.

ben-manes:
We delegate to Gradle for resolution, which should be looking at the `maven-metadata.xml` to find the version candidates. Which repositories are you using? I recall jcenter used to bad results (mix of new and stale) which could confuse Gradle's dependency resolver.

westito:
I use these:
```
google()
jcenter()
mavenCentral()
maven { url "https://jitpack.io" }
```

I tried with a single repository (only jcenter or mavenCentral) and cleared gradle / Studio cache, but the result is the same. 

Gradle:  v8.2.1 (tried v8.0.2 too)
com.android.tools.build:gradle:8.1.0 (tried 8.0.2 too)

Gradle logs this XMLs when checking version:
```
https://repo.maven.apache.org/maven2/com/squareup/okhttp3/okhttp/maven-metadata.xml
https://repo.maven.apache.org/maven2/com/squareup/okhttp3/okhttp/4.10.0/okhttp-4.10.0.pom / module
https://repo.maven.apache.org/maven2/com/squareup/okhttp3/okhttp/4.9.3/okhttp-4.9.3.pom / module
https://repo.maven.apache.org/maven2/com/squareup/okhttp3/okhttp/5.0.0-alpha [ lot of alphas ]
```

westito:
Linter reports the right version in build.gradle file inspection: `A newer version of com.squareup.okhttp3:okhttp than 4.9.3 is available: 4.11.0`

ben-manes:
I wrote a minimal example which works,

```gradle
plugins {
  id 'com.github.ben-manes.versions' version '0.47.0'
  id 'java-library'
}

repositories {
[...]

Issue #550, opened 2021-07-28
SLF4J milestones returned with returned with -Drevision=release


Thunderforge:
When running `gradlew dependencyUpdates -Drevision=release`, SLF4J API milestones are included (e.g. [2.0.0-alpha2](https://mvnrepository.com/artifact/org.slf4j/slf4j-api/2.0.0-alpha2)). According to [this section of the docs](https://github.com/ben-manes/gradle-versions-plugin#revisions), this should only be shown when `-Drevision=milestone`.

## Expected Behavior

* `gradlew dependencyUpdates -Drevision=release` results in  `org.slf4j:slf4j-api [1.7.29 -> 1.7.30]`
* `gradlew dependencyUpdates -Drevision=milestone` results in  `org.slf4j:slf4j-api [1.7.29 -> 2.0.0-alpha2]`

## Actual Behavior

* `gradlew dependencyUpdates -Drevision=release` results in  `org.slf4j:slf4j-api [1.7.29 -> 2.0.0-alpha2]`
* `gradlew dependencyUpdates -Drevision=milestone` results in  `org.slf4j:slf4j-api [1.7.29 -> 2.0.0-alpha2]`

ben-manes:
`revision` is an Apache Ivy concept, which Gradle was originally built upon. At that time, this was the only flag to control resolution. As few use `ivy.xml` metadata, it does not match to custom versioning conventions. In the readme it warns as,

>The revision task property controls the Ivy resolution strategy for determining what constitutes the latest version of a dependency. Maven's dependency metadata does not distinguish between milestone and release versions. The following strategies are natively supported by Gradle:
>
>release: selects the latest release
>milestone: select the latest version being either a milestone or a release (default)
>integration: selects the latest revision of the dependency module (such as SNAPSHOT)

In your case, you would want to use a `resolutionStrategy`, such as `rejectVersionIf`. This allows fine-grained control over versioning which falls outside of Maven's conventions.

Thunderforge:
> In your case, you would want to use a `resolutionStrategy`, such as `rejectVersionIf`. This allows fine-grained control over versioning which falls outside of Maven's conventions.

Can this plugin be modified to just do this instead of leaving it up to the user?

[...]

Issue #440, opened 2020-11-03
Better default that does not show "-rc" and "-alpha" versions


lathspell:
It would be nice if the default would be to only show updates for "release" class dependencies. Currently the plugin shows, without further configuration, updates like the following which clearly have "alpha" or "rc" in their versons:

* ch.qos.logback:logback-classic [1.2.3 -> 1.3.0-alpha5]
* com.fasterxml.jackson.module:jackson-module-kotlin [2.11.3 -> 2.12.0-rc1]
* org.jetbrains.kotlin:kotlin-annotation-processing-gradle [1.4.10 -> 1.4.20-RC]
* org.slf4j:slf4j-api [1.7.30 -> 2.0.0-alpha1]

I read about the possibility of tweaking the plugin with "rejectVersionIf" but I don't like cluttering every Gradle file with additional functions and filtering "alpha" and "rc" when "revsion=release" is active really should be the default.

ben-manes:
`revsion` is an Apache Ivy concept which early versions of Gradle was built on top of. Back then it was the only toggle we had. This only removes snapshots from Maven dependencies because that is their versioning convention. The alpha, rc, etc is considered a stable release version by Maven. This flag is mostly historical and isn’t useful in practice.

You can share build scripts across independent builds, e.g. init scripts. This would avoid your duplication problem.

lathspell:
My concern was more if the current plugin defaults are sensible? I would guess that most users would usually not want to see alpha and rc versions for their dependencies and consider them as "false-positives". So it would be better if they would be hidden by default.

ben-manes:
Unfortunately since there is no common convention, I’m afraid if we impose one then it will cause more confusion. The only standard is governed by the repository type, Maven or Ivy, and that is what Gradle applies by default. They then offer configuration knobs to tune those results, so we expose them.

For example some use even/odd versions to designate release and beta versions. Others use keywords, which vary considerably. Some use keywords after a release to indicate patches, e.g. 1.0-SR3 as the 3rd patch after the 1.0-RELEASE version. Some use keywords to split target platforms, e.g. -jre vs -android. Others use timestamp version strings and hack means to backport patches between major releases.

It may appear lazy for us to not provide a sane set of defaults, but the patterns we see in the wild are insane. It seems less confusing therefore to match expectations from Gradle and expose its configuration tools to let users tune it as needed.

[...]

PR #1093, opened 2026-09-04
Reject pre-release candidates by default with a `rejectPreReleases` task property

ianbrandt:
I've been looking into #440, and I've prototyped an approach that might be worth considering.

The [`rejectVersionIf`/`isNonStable` recipe](https://github.com/ben-manes/gradle-versions-plugin#filtering-unstable-versions) in the README matches a pattern for stable versions, so it filters out any release with a qualifier the pattern doesn't cover, such as `13.4.0.jre11`, `1.1.17.SP2`, `9.2-1002-jdbc4` and every Spring `.SECnn` security release.

If we flip this around, and only filter well-known pre-release version strings and versions ending in a commit hash, I believe we can get a much lower error rate. A misdetection is then simply passed through to the report, as it is today with no `rejectVersionIf` configured, and the user can add custom handling for such patterns if they choose.

This PR adds that as a `rejectPreReleases` task property, on by default, which leaves a pre-release candidate out of the report unless the current version is itself a pre-release. Measured over the entire published version history of 1,870 artifacts on Maven Central, Google's Maven repository, Clojars and the Gradle Plugin Portal, 28,735 distinct version strings:

| | README `isNonStable` recipe | `rejectPreReleases = true` |
| --- | --- | --- |
| Versions filtered out | 41.2% | 36.9% |
| Releases wrongly filtered out | 873 (3.0%) | 0 |
| Pre-releases missed | 1 (0.0%) | 375 (1.3%) |

Two dependencies as reported today, with nothing configured:

```
 - com.microsoft.sqlserver:mssql-jdbc [10.2.0.jre11 -> 13.5.1.jre11-preview]
     https://github.com/Microsoft/mssql-jdbc
 - org.slf4j:slf4j-api [1.7.29 -> 2.1.0-alpha1]
     http://www.slf4j.org
```

The same build with the recipe. slf4j is corrected but the preview is not, because `10.2.0.jre11` matches as a pre-release too, so the `&& !isNonStable(currentVersion)` clause never fires:

```
 - com.microsoft.sqlserver:mssql-jdbc [10.2.0.jre11 -> 13.5.1.jre11-preview]
     https://github.com/Microsoft/mssql-jdbc
 - org.slf4j:slf4j-api [1.7.29 -> 2.0.19]
     http://www.slf4j.org
```

The same build on this branch, again with nothing configured:

```
 - com.microsoft.sqlserver:mssql-jdbc [10.2.0.jre11 -> 13.4.0.jre11]
     https://github.com/Microsoft/mssql-jdbc
 - org.slf4j:slf4j-api [1.7.29 -> 2.0.19]
     http://www.slf4j.org
```

A convention the built-in markers don't cover, such as graphql-java's `-nf-` builds, is added to the check with `preReleaseVersionIf`, so the property and its option govern it too:

```kotlin
preReleaseVersionIf { it.contains("-nf-") }
```

A module is exempted from the built-in checks with `exemptFromBuiltInChecksIf`. It takes the same predicate as `rejectVersionIf` and is applied inside the checks, so the properties stay on for the rest of the build and the options apply as they do without it:

```kotlin
exemptFromBuiltInChecksIf { candidate.module == "guava" }
```

`rejectVersionIf` is unchanged and is applied in addition, for policy that isn't about pre-releases or bounds. Both checks are readable from a rule, as `isPreRelease()`, `isPreRelease(version)` and `isOutOfDeclaredBounds()`, so a policy of your own can build on them.

On the survey: I ran both approaches over it with the help of AI. Central's 1,005 artifacts came from ecosyste.ms ranked by dependent package count, and Clojars' 300 from its published all-time download stats. No ranking is published for the portal, so its 492 came from a hand-written list of well-known plugins plus a keyword search. AI flagged 91 of the filtered versions as possible real releases, but on inspection none of them was.

It's on by default, as `rejectOutOfBounds` from #1092 is, since I believe that covers the majority use case with the least configuration. It's off by default under the `integration` revision, and off for a run with `--no-reject-pre-releases`.

This also renames `rejectOutOfBoundVersions` from #1092 to `rejectOutOfBounds`, since the plugin reports on versions and the suffix restated what every property of the task is about. Neither name has been released, so no build compiled against a published jar is affected.

This also relates to #550 and #798.


ben-manes (approved):
I am okay since you're driving the development efforts now and it becomes your call wrt maintainability. I am a little hesitant due to how chaotic it was historically that I did not want to dive into what defined a pre-release, and instead tried to deflect that by giving users the tools to decide. In those days weird schemes like "odd version numbers are beta" existed, and Maven's own version guide was only loosely followed at best. These days with semver there is a clearer understanding of what is proper.

Instead of many flags, you might consider adding methods to `ResolutionStrategyWithCurrent` for a DSL to make it just as easy but still defer to the user to configure  for pre-canned templates. Then its still some configuration but you avoid a matrix of options that could intersect poorly, or many top-level flags, so it reads behaviorally.

```kts
rejectVersionIf { isPreRelease(candidate.version) && !isPreRelease(currentVersion) }
// or combined
rejectVersionIf { becomesUnsable()  }
```

ianbrandt:
Roger all. I initially wasn't sure I was going to end up proposing this, as I was familiar with your past comments expressing hesitancy towards the approach. Then I saw the numbers for how the fail-open strategy with common SemVer patterns works out for the vast majority of the most-popular/depended-on artifacts. For that corpus, passing through 1.3% of pre-releases to the report instead of wrongly filtering 3.0% stable releases with the stock `isNonStable` recipe is what sold me.

One thing I didn't research is how often users of this plugin take the `isNonStable` recipe and apply it stock. That's what I've always done, but anecdote isn't data. If any significant percentage of others do apply it unmodified, the case could be made that it's boilerplate. That would be another pro for this approach. Now that I've thought about it, I'm curious. I'll do the research and report back.

The reason I thought task options would be good for this is it allows one to track stable releases by default for reliable builds, but then also one-shot see what pre-releases are available for testing with `--no-reject-pre-releases`. With the DSL approach, I don't believe we'd get that use case covered for free like with task options. That said, poorly interacting options and too many top-level toggles wouldn't make for great user experience either. I'll think on this, and see if I can't come up with a best-of-both or 3rd alternative.

One thing I haven't mentioned is that I was led down this #1092 and #1093 road by another PR I've been working on to hopefully further improve support for composite builds. I'll follow up with a summary of what I'm up to there, and how convention over configuration for `rejectOutOfBounds` and `rejectPreReleases` (whether default on or off) supports it.

ianbrandt:
I've thought about the DSL some more, and I think it makes sense in addition to the task properties. The DSL supports composition, and the properties support a command-line override of the current setting or default. With `--no-reject-pre-releases` the pre-releases are reported for a single run with no edit to the build script. There's no command-line equivalent for a rule in `rejectVersionIf`. With both properties on by default there's nothing to configure in the common case.

I've added the DSL to this PR, on `ComponentSelectionWithCurrent`: `isPreRelease()` for the check the property applies, true when the candidate is a pre-release and the current version is not, `isPreRelease(version)` for the version-level test behind it, and `isOutOfDeclaredBounds()` beside them. A lambda that only calls these members can be serialized, which matters for the composite build PR I'm working on (see below).

For a convention the markers miss, like graphql-java's `-nf-` builds, I've also added `preReleaseVersionIf { it.contains("-nf-") }`. That extends what counts as a pre-release rather than adding a rejection rule, so the same definition is applied by the built-in check, by the exemption for a build already on a pre-release, and by `isPreRelease` in a rule. It also means that with `--no-reject-pre-releases` the built-in check and the added convention are both off for that run. For an exception, I've added `exemptFromBuiltInChecksIf { candidate.module == "guava" }`. It takes the same predicate as `rejectVersionIf`, and it's applied inside the two built-in checks, so the checks stay on for the rest of the build and the properties and options apply as they do without it. `rejectVersionIf` stays for policy that isn't about pre-releases or bounds, like pinning a module.

Regarding a matrix of options that could intersect poorly, the only one that stands out to me is that I have `rejectPreReleases` defaulting to off if `revision = "integration"`. Since `"integration"` selects the newest version whatever its status, snapshots included, a pre-release filter would contradict it. It's the one place a setting changes another's default. Everything else composes without any design integrity issues as far as I can tell. Let me know if you see any others.

On the `isNonStable` research, I took the population from GitHub's code search: repos with `"com.github.ben-manes.versions"` in a `.gradle`, `.kts` or `.toml` file. I checked each one for a version filter, and classified every filter I found against the 15 recipe texts across the README's 181 revisions:

* About 30% of the repos that apply the plugin filter versions at all. About 68% of those use a copy of the README recipe rather than a hand-written policy.
* Nearly every filter I found was filtering pre-releases. Around 5% do something else, mostly pinning a module.
* Both hold per owner as well as per repo, so neither figure comes from fleets of related repos or from forks.

The majority (~70%) of the builds in the corpus don't filter at all. I don't think that's evidence either way on whether most authors want pre-releases filtered by default. We at least have the related issue requests. Also, the corpus is all public GitHub repos. I could see private commercial repos being more likely to filter pre-releases for the sake of reliable builds. That said, I concede leaving pre-releases out by default is a judgment call.

Regarding my incubating composite build PR, today the rows from an included build are filtered only by what is configured in that build. In that PR the full candidate list is recorded per project, and the rules on the aggregating build's `dependencyUpdates` task are applied to every row of the merged report, from subprojects and included builds alike. A rule has to be serialized into the configuration cache for that. The Kotlin `isNonStable` recipe as written in the README doesn't serialize, since the lambda calls a helper defined in the script and so captures the script. Moving the helper into build logic fixes it, but with the addition of `rejectOutOfBounds` and `rejectPreReleases`, the thinking is most builds won't need a rule at all, eliminating the need to migrate the `isNonStable` recipe to build logic in order to take up the planned composite build improvement.

ianbrandt:
Merged after some further improvements. Composite build PR posted: #1094.

PR #1094, opened 2026-09-08
Apply the report's settings to the entries merged from an included build

ianbrandt:
This PR applies the settings that control what a report shows to every entry merged into it, so a composite build is configured in one place, as a multi-project build is.

Today configuring the root project's `dependencyUpdates` task covers every subproject, since the task settings are inherited within a build, but none of them are applied to an included build merged into that report through `dependencyUpdatesAggregation`. Its entries are reported as its own resolution accepted them, so a `rejectVersionIf` on the aggregating task is never applied to them. The way to get one consistent report is to copy the task configuration into every included build.

On this branch each `partialDependencyUpdates` task records every candidate listed in the repositories, and the `dependencyUpdates` task that merges the partials applies its own `rejectVersionIf`, `resolutionStrategy`, `rejectPreReleases`, `preReleaseVersionIf`, `exemptFromBuiltInChecksIf` and `filterDeclaredConfigurations` to what it merges.

A composite with the aggregation declared in the including build, and a rule there rejecting a version that resolution in the included build accepted:

```kotlin
dependencies {
  dependencyUpdatesAggregation("com.example:child:1.0")
}

tasks.dependencyUpdates {
  rejectVersionIf { candidate.version == "3.1" }
}
```

The merged report today shows the version resolved in the included build:

```
 - com.google.inject:guice [2.0 -> 3.1]
```

The same build on this branch:

```
 - com.google.inject:guice [2.0 -> 3.0]
```

Within one build, the report's rule is now applied over a subproject's own as well. Where a rule on the project the report is asked for rejects a version a subproject's rule accepts, one entry at the older version is shown for every project, where a per-project rule split them into two before.

Applying the rules at the report only leaves candidates out. An entry can be moved to an older version and never to a newer one, and the older version was never resolved, since the candidates come from a repository listing. Every project of the build declared by a `dependencyUpdatesAggregation` entry is now merged, where only the project its coordinates resolved to was merged before, and an entry that Gradle substitutes onto no project is printed in a warning rather than dropped.

The settings that control what is resolved, `revision`, `filterConfigurations`, `checkConstraints`, `checkBuildEnvironmentConstraints` and `rejectOutOfBounds`, are still read in the build that resolves, and are declared in each included build.

Where a report's rules are applied to another build's entries, they are stored in the configuration cache. A Kotlin rule that calls a function declared in the same build script captures the script, which the cache cannot store, so the entry is discarded and the project is printed in a warning. Moving the function into `buildSrc` or an included build keeps the entry, and a Groovy closure is unaffected. With `rejectOutOfBounds` from #1092 and `rejectPreReleases` from #1093 on by default, most builds need no rule at all, and a rule that only calls `isPreRelease()` or `isOutOfDeclaredBounds()` keeps the entry too. The partial file format is bumped, so a report that merges a partial written by a newer plugin version fails with the project named rather than reading it wrong.

The README's [Shared task settings](https://github.com/ben-manes/gradle-versions-plugin#shared-task-settings) and [Composite builds](https://github.com/ben-manes/gradle-versions-plugin#composite-builds) sections and the migration notes are updated with the above. Nothing in the report model handed to `outputFormatter` is changed.


gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/resolutionstrategy/RecordedComponentSelection.kt
```diff
@@ -0,0 +1,118 @@
...
+    if (this === other) return true
+    if (other !is RecordedModuleComponentIdentifier) return false
+    return group == other.group && module == other.module && version == other.version
+  }
+
+  override fun hashCode(): Int {
+    var result = group.hashCode()
+    result = 31 * result + module.hashCode()
+    result = 31 * result + version.hashCode()
+    return result
```
ben-manes: fwiw,
`return Objects.hash(group, module, version)`

ben-manes (approved):

gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/resolutionstrategy/RecordedComponentSelection.kt
ianbrandt: Thanks. Done. I also updated `Coordinate.hashCode`, which was the only other hand-rolled instance I found.

PR #1096, opened 2026-09-10
Exempt the version lookups from dependency verification (fixes #1095)

### LINKED ISSUE #1095: `dependencyUpdates` fails on Gradle 8.7 and later when dependency verification is enabled
On Gradle 8.7 and later, `dependencyUpdates` fails when the build is set up for dependency verification with `<verify-metadata>true</verify-metadata>`. The candidate version's pom cannot be in the verification metadata, because that is the newer version being searched for.

Reproducer, on 0.61.0 with nothing else in the build:

```groovy
plugins {
  id 'java'
  id 'io.github.ben-manes.versions' version '0.61.0'
}

repositories {
  mavenCentral()
}

dependencies {
  implementation 'com.google.inject:guice:5.1.0'
}
```

```
$ gradle --write-verification-metadata sha256 build
$ gradle dependencyUpdates

* What went wrong:
Execution failed for task ':dependencyUpdates'.
> Dependency verification failed for configuration ':detachedConfiguration1'
  2 artifacts failed verification:
    - guice-7.0.0.pom (com.google.inject:guice:7.0.0) from repository MavenRepo
    - guice-parent-7.0.0.pom (com.google.inject:guice-parent:7.0.0) from repository MavenRepo
```

* On Gradle 8.4, 8.5 and 8.6 the task succeeds, with ` - com.google.inject:guice [5.1.0 -> 7.0.0]` in the report. I get the failure on 8.7 through 8.14.4 and on 9.7.1.
* Setting `<verify-metadata>false</verify-metadata>` works around it, at the cost of metadata verification for the whole build.
* I have a fix and will post a PR.

ianbrandt:
Fixes #1095

This PR exempts the plugin's version lookups from the build's dependency verification. The three query copies opt out through their resolution strategy, and the project url lookup resolves its pom through a detached configuration, since an `ArtifactResolutionQuery` has none to opt out with.

One user-visible consequence: where an included build substitutes a module, the report now omits that module's project url, because the resolved component is the local project rather than the published module. The first commit also corrects three sha256 values in the existing verification test, which were taken from Maven Central's real poms rather than the trimmed copies in the test resources. The regression test runs on 8.6, 8.7, 8.14.4 and 9.7.1, where 8.6 is the last release that passes without the exemption.


gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/Resolver.kt
```diff
@@ -910,34 +917,49 @@ class Resolver internal constructor(
...
-          if (artifact is ResolvedArtifactResult) {
-            val file = artifact.file
-            project.logger.info("Pom file for $id is $file")
-            var url = interpolate(getUrlFromPom(file), id)
+      // https://github.com/ben-manes/gradle-versions-plugin/issues/1095
+      // An ArtifactResolutionQuery cannot be exempted from dependency verification, as there is
+      // no resolution strategy on it, and the candidate version is never in the build's metadata.
+      // The artifact-only notation is what keeps an included build that substitutes the module
+      // from being built to produce its jar. The coordinates are passed as fields rather than
+      // interpolated into a string, which an '@' in the version would be read as splitting.
```
ben-manes, line 925: The docs for `ArtifactResolutionQuery` also warn it will later be deprecated for removal, so this is better than doubling down on it anyway

gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/Resolver.kt
```diff
@@ -910,34 +917,49 @@ class Resolver internal constructor(
...
+
+      // empty for gradle plugins, a single pom for normal dependencies
+      for (artifact in artifacts) {
+        val file = artifact.file
+        project.logger.info("Pom file for $id is $file")
+        var url = interpolate(getUrlFromPom(file), id)
+        if (!url.isNullOrEmpty()) {
+          project.logger.info("Found url for $id: $url")
+          return url.trim()
+        } else {
```
ben-manes: PMD has beaten it into my head of `if (!false) ... else /* true */` 
should be rewritten as `if (true) ... else ...`

ben-manes (approved):

gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/Resolver.kt
ianbrandt: Agreed. The `if` returns, so I dropped the `else` rather than inverting it. That takes out the negation-with-else and a level of nesting.

gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/Resolver.kt
ianbrandt: Good to know!

Issue #69, opened 2015-07-02
Show version breadcrumbs


ben-manes:
@tkruse suggestion is to display the version upgrade path, e.g.

``` bash
- org.mockito:mockito-all [1.10.18 -> 1.10.19 -> 1.11.5 -> 2.1.6]
```

> meaning the latest version with same major and minor, the latest with same major, and the global latest. That would serve the same purpose, but go without additional configuration.

The trick for the expanded version list is getting it through the Gradle API. Gradle expects to select an exact version or the latest that meets a criteria. We could have a rule that rejects all versions (except for the current) and captures the version list. This would be a roundabout way to get the information, but would let us show the transition flow.

jochenberger:
Wouldn't we have to do multiple resolution steps to get that information, i.e. one for `1.10.+`, one for `1.+` and one for `+`?

ben-manes:
Yep, it would be multiple and we'd have to do it using a `ComponentSelection` rule that captures the candidate's version and rejects until it reaches the build's current version. The more out of date the dependency is, the more versions have to be resolved and the slower the task takes.

I'm pretty confident we can get this to work, but I'm not sure if most users would appreciate the enhancement or prefer a faster check. @tkruse idea was that this approach would replace users adding a custom component selection rule because they could filter out betas by inspection.

What do you think we should do?

tkruse:
As long as multiple calls are required, I would not want this as standard behavior. I am not sure how easy or convenient it is to add some switch such that users may run with '--verbose' or something to get more information

ben-manes:
Unfortunately it would require multiple calls because Gradle doesn't expose to us the version metadata (e.g. from maven_metadata.xml). It appears that when evaluating candidates it retrieves the candidate version's metadata (e.g. pom.xml) so an extra http request is required per candidate evaluation. This is then cached, but due to the sequential nature of the resolution process this is slow.

Its probably time to bite the bullet and add a configuration block (dsl extension). I had avoided that in the past because I didn't want a dozen toggles to hack around api quirks. I think the ones we'd need now all reasonably fit into Gradle's api instead of working around it. I'm thinking we'd want options like:
 - breadcrumbs: boolean to display version path (is there a better name?)
 - resolutionStrategy: block for rules only applied to the task's copied configuration
 - versionPattern: A regex string for the version as a one-line alternative to the block
 - revision: The revision filter (integration, milestone, release)
 - misc reporting: #12, #70, formatters, etc. 

[...]

PR #1097, opened 2026-09-11
Report the latest pre-release after the latest release by default

ianbrandt:
I realized that with the new pre-release filtering logic from #1093 we can print
the same `[current -> stable -> pre-release]` breadcrumb for dependency updates
that we already print for Gradle updates. That also flipped my thinking on the
default value for `rejectPreReleases`. In this PR it defaults to `false` to align
with the released default of no pre-release filtering and `release-candidate` for
`gradleReleaseChannel`.

`rejectPreReleases` is now about what is printed rather than about which
candidate is resolved. At the new default, both the latest stable release and any
later pre-release are printed. Set it to `true` and only the latest stable
release is printed.

`preReleaseVersionIf` from #1093 is still available to add pre-release logic
covering versions such as graphql-java's `-nf-` builds. A version it
matches is printed as the later pre-release, rather than left out of the report.

The default for `gradleReleaseChannel` now follows `rejectPreReleases`. At the
new default it stays `release-candidate`, so the Gradle release candidate is
still printed. Set `rejectPreReleases = true` and it becomes `current`, so the
release candidate is not. Setting `gradleReleaseChannel` in the build, passing
`--gradle-release-channel`, or setting the system property is read ahead of that.
`--reject-pre-releases` and `--no-reject-pre-releases` apply to both for one run.

On `master` today, with the defaults, a release candidate is printed for Gradle
and left out for a dependency:

```
 - com.google.guava:guava [32.0.0-jre -> 33.7.1-jre]
 - org.slf4j:slf4j-api [2.0.9 -> 2.0.19]

 - Gradle: [8.4 -> 9.7.1 -> 9.8.0-rc-1]
```

On `master` today, with `rejectPreReleases = false`, the 2.0.19 release is lost:

```
 - org.slf4j:slf4j-api [2.0.9 -> 2.1.0-alpha1]
```

On this branch, with the defaults:

```
 - com.google.guava:guava [32.0.0-jre -> 33.7.1-jre]
 - org.slf4j:slf4j-api [2.0.9 -> 2.0.19 -> 2.1.0-alpha1]

 - Gradle: [8.4 -> 9.7.1 -> 9.8.0-rc-1]
```

On this branch, with `rejectPreReleases = true`:

```
 - org.slf4j:slf4j-api [2.0.9 -> 2.0.19]

Gradle current updates:
 - Gradle: [8.4 -> 9.7.1]
```

Each step is printed only where it is later than the one before it, so a module
with no later stable release is printed as `[2.0.9 -> 2.1.0-alpha1]`.

Details:

- The pre-release step is the latest candidate that fails the pre-release check
  alone. The bound check, the revision and any `rejectVersionIf` filter are
  applied first, so a candidate already rejected by one of those is not printed.
- The step is in `available.preRelease` in the JSON and XML reports. Where there
  is no later stable release, `available.release`, `available.milestone` and
  `available.integration` are all null.
- A `rejectVersionIf` filter is now called for pre-release candidates it was
  never called for before, since the built-in check is applied last. If your
  filter has a side effect, it now happens for those candidates too.
- I bumped the partial report format to 3. Read by an older release, a partial
  written by a later one would have its `latestVersion` taken under the earlier
  meaning, and the bump makes that combination fail instead.
- I added a fourth constructor argument to `VersionAvailable`. Every arity
  shipped in the last release is still callable, so Groovy and Java callers are
  unaffected, but Kotlin code constructing one with named arguments needs
  recompiling.

### DIFF
--- a/README.md
+++ b/README.md
@@ -308,14 +308,14 @@ tasks.named("dependencyUpdates").configure {
 - `checkConstraints` adds the versions a `constraints` block manages to the
   report (see [Constraints](#constraints)).
 
-Nothing else is needed. A pre-release candidate is left out by
-`rejectPreReleases`, and a candidate outside a `strictly` or `reject` bound
-written in the build, outside a dynamic version declared on the buildscript
-classpath, or outside the version fixed by a consumed platform is left out by
-`rejectOutOfBounds`. Both are on by default (see [Filtering unstable
+Nothing else is needed. A pre-release candidate is printed as a step of its own
+after the newest release rather than in place of it, and `rejectPreReleases`
+leaves that step out. A candidate outside a `strictly` or `reject` bound written
+in the build, outside a dynamic version declared on the buildscript classpath, or
+outside the version fixed by a consumed platform is left out by
+`rejectOutOfBounds`, which is on by default (see [Filtering unstable
 versions](#filtering-unstable-versions) and [Respecting declared
-bounds](#respecting-declared-bounds)); set either to `false` to see what it
-leaves out.
+bounds](#respecting-declared-bounds)).
 
 The [configuration filters](#configuration-filter) are absent only because
 their arguments are build-specific: the names to reject come from your own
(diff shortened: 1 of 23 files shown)


ben-manes:
oh, take a look at https://github.com/ben-manes/gradle-versions-plugin/issues/69 to see if there are other ideas you want to extend this with.

ben-manes (approved):

ianbrandt:
> oh, take a look at #69 to see if there are other ideas you want to extend this with.

Thanks for the pointer! The resolution per step concern is no longer an issue since #1094. Every candidate version is already recorded per project, with no metadata read, so the same-major and same-minor steps can be determined with what we already have. We just need to do the SemVer parsing. I'll look into a separate PR for it.

Issue #732, opened 2023-02-22
Unable to find method


sdoward:
I see the follow exception when running `dependencyUpdates`.

```
Execution failed for task ':dependencyUpdates'.
> 'okio.Buffer okio.BufferedSource.getBuffer()'
```

Version: 0.46.0
Gradle Version: 7.4.2

ben-manes:
Probably a version conflict between plugins. You can run buildEnvironment to see if our transitive version was changed to an incompatible release.

https://github.com/ben-manes/gradle-versions-plugin/blob/86cb0eebbfa3986c3e93c5f243c08d51882367d4/gradle-versions-plugin/build.gradle#L34-L35

sdoward:
@ben-manes thanks for the quick response.

Actually I only see this plugin with those dependencies...

+--- net.gpedro.integrations.slack:slack-webhook:1.2.1
|    \--- com.google.code.gson:gson:2.3.1 -> 2.8.6
+--- com.android.tools.build:gradle:7.3.1
|    +--- com.android.tools:sdk-common:30.3.1
|    |    +--- com.android.tools.analytics-library:shared:30.3.1
|    |    |    +--- com.android.tools.analytics-library:protos:30.3.1
|    |    |    |    \--- com.google.protobuf:protobuf-java:3.17.2
|    |    |    +--- com.android.tools:annotations:30.3.1
|    |    |    +--- com.android.tools:common:30.3.1
|    |    |    |    +--- com.android.tools:annotations:30.3.1
|    |    |    |    +--- com.google.guava:guava:30.1-jre
|    |    |    |    |    +--- com.google.guava:failureaccess:1.0.1
|    |    |    |    |    +--- com.google.guava:listenablefuture:9999.0-empty-to-avoid-conflict-with-guava
|    |    |    |    |    +--- com.google.code.findbugs:jsr305:3.0.2
|    |    |    |    |    +--- org.checkerframework:checker-qual:3.5.0
|    |    |    |    |    +--- com.google.errorprone:error_prone_annotations:2.3.4 -> 2.4.0
|    |    |    |    |    \--- com.google.j2objc:j2objc-annotations:1.3
|    |    |    |    +--- net.java.dev.jna:jna-platform:5.6.0
|    |    |    |    |    \--- net.java.dev.jna:jna:5.6.0
|    |    |    |    \--- org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.5.31
[...]

PR #1120, opened 2026-09-15
Time out the Gradle update check after 15s (fixes #1119)

### LINKED ISSUE #1119: A stalled Gradle version lookup can block `dependencyUpdates` for hours
The timeouts of the Gradle update check are 15,000 seconds, about four hours. `CLIENT_TIME_OUT` is 15,000, and it's passed to OkHttp with `TimeUnit.SECONDS`:

https://github.com/ben-manes/gradle-versions-plugin/blob/v0.62.0/gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/gradle/GradleUpdateChecker.kt#L88-L94

* With `gradleVersionsApiBaseUrl` pointed at a local server that accepts the request and never responds, `GradleUpdateChecker` was still waiting after 60 seconds.
* The 15 second timeout was added for #241 in v0.20.0, and the unit changed with the move to OkHttp in v0.43.0.

The update check is also the only use of OkHttp, for three GET requests. OkHttp is on the buildscript classpath of every build that applies the plugin:

```
classpath
+--- com.github.ben-manes.versions:com.github.ben-manes.versions.gradle.plugin:0.62.0
|    \--- io.github.ben-manes:gradle-versions-plugin:0.62.0
...
|         +--- com.squareup.okhttp3:okhttp:4.12.0
|         |    +--- com.squareup.okio:okio:3.6.0
...
|         \--- com.squareup.moshi:moshi-kotlin:1.12.0
```

* A version skew between OkHttp and Okio from another plugin broke `dependencyUpdates` in #732.
* Dependabot has opened a bump to OkHttp 5 five times: #952, #954, #971, #1015, and #1114.
* OkHttp 5.5.0 depends on kotlin-stdlib 2.1.21, and Gradle 8.4, our minimum, embeds Kotlin 1.9.10.
* `HttpURLConnection` is available on JDK 8 and covers the three requests.
* I have a fix for both and will post a PR.

Verified on v0.62.0.

ianbrandt:
Fixes #1119

This PR fixes the unit of the Gradle update check timeout, which restores the 15s added for #241. A version lookup that never responds now fails after 15s. The timeout isn't currently user-configurable, but a task property could be added for it if anyone ever asks for that.

For testing, an internal `GradleUpdateChecker` constructor that takes a timeout was added so the `GradleUpdateCheckerSpec` test for timeout handling doesn't wait 15s in each test task.

OkHttp is replaced with `HttpURLConnection`, and is no longer on the buildscript classpath of builds that apply the plugin. Without OkHttp, there is one less library to conflict with other plugins, as in #732, and one less to keep current, as with the OkHttp 5 bump in #1114. Okio stays because Moshi depends on it, and Moshi is bumped to 1.15.2 so that Okio is at 3.7.0, past the fix for CVE-2023-3635. Credentials from a registered `java.net.Authenticator` are now sent to a host that responds with a 401 or a proxy that responds with a 407, and a redirect to another scheme is no longer followed.


gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/gradle/GradleUpdateChecker.kt
```diff
@@ -81,48 +86,44 @@ class GradleUpdateChecker(
...
-            response.body?.source()?.let { body ->
-              val version = moshi.adapter(VersionSite::class.java).fromJson(body)?.version.orEmpty()
-              if (version.isNotEmpty()) {
-                cacheMap[it] = ReleaseStatus.Available(GradleVersion.version(version))
-              } else {
-                cacheMap[it] = ReleaseStatus.Unavailable
-              }
+          val connection = URI(gradleVersionsApiBaseUrl + it.id).toURL().openConnection() as HttpURLConnection
+          connection.connectTimeout = timeoutMillis
+          connection.readTimeout = timeoutMillis
```
ben-manes, line 110: it may not be any cleaner, but fwiw HttpClient is likely available in the minimum JVM required by Gradle.

ben-manes (approved):

gradle-versions-plugin/src/main/kotlin/com/github/benmanes/gradle/versions/updates/gradle/GradleUpdateChecker.kt
ianbrandt: I looked into that. `HttpClient` was added in Java 11, so we'd have to drop our Java 8-10 support to switch. That didn't seem worth it for this minimal use case. That said, let me know if you want me to bump it up. We could keep the Gradle 8.4 floor.

ben-manes: its fine with me either way, and this was reasonable as is. I haven't used the client much, so no preference

ianbrandt: Very good.

Merging this and prepping the v0.63.0 release with the recent changes. A snapshot based on this PR is looking good in all my local testing.

